PR #14, opened 2020-05-19
Introduce ContentSource and ContentLibraryProvider CRDs

aruneshpa:
In order to make virtual machine images available to a developer, a VI admin exposes them by adding them to a content library and configuring a cluster to use that content library. This configuration is done by plugging in the content library uuid in VM operator's provider ConfigMap.

This presents a few challenges:
- Only one content library can be associated with a VM operator instance
- Image discovery has to be driven via the images controller. While it works, this should be driven from top down. A content library should create the images in that library

This change introduces new CRDs to enhance the image support in VM operator.

### ContentSource
A ContentSource describes a content source. This resources contains a reference to a resource which describes a content provider. A content library is one such example of a content provider. Content providers have their owner references set to the ContentSource.

**Example ContentSource resource**

```
apiVersion: vmoperator.vmware.com/v1alpha1
kind: ContentSource
metadata:
  name: parunesh-contentsource
spec:
  providerRef:
    apiVersion: vmoperator.vmware.com/v1alpha1
    kind: ContentLibraryProvider
    name: parunesh-contentlib-vmtx
```

### ContentLibraryProvider
A ContentLibraryProvider is a content provider which uniquely describes a vSphere content library. Currently, only the content library UUID is used to identify a library on vSphere.

**Example ContentLibraryProvider resource**
```
apiVersion: vmoperator.vmware.com/v1alpha1
kind: ContentLibraryProvider
metadata:
  name: parunesh-contentlib-vmtx
ownerReferences:
  apiVersion: vmoperator.vmware.com/v1alpha1
  kind: ContentSource
  name: parunesh-contentsource
  uid: 7bbd5927-fd57-4e8c-adae-8fa162cf80e8
spec:
  uuid: e5d287d1-a6e0-48fb-b6f0-34ca0996729b
```


api/v1alpha1/contentlibraryprovider_types.go
```diff
@@ -0,0 +1,48 @@
...
+	UUID string `json:"uuid,omitempty"`
+}
+
+// ContentLibraryProviderStatus defines the observed state of ContentLibraryProvider
+// Can include fields indicating when was the last time VM images were updated from a library
+type ContentLibraryProviderStatus struct {
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Cluster
```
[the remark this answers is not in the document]
aruneshpa: Good point. Let me add the columns.
Instead of `CL`, how about we call it either `Content Library` or `Library`.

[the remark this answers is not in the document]
janitha: That works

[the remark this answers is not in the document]
aruneshpa: Added the custom printcolumns. This is how the output looks like:

```
NAME                       CONTENT LIBRARY NAME   CONTENT LIBRARY UUID
parunesh-contentlib-vmtx   vsphere name           e5d287d1-a6e0-48fb-b6f0-34ca0996729b
```

api/v1alpha1/contentlibraryprovider_types.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ContentLibraryProviderSpec defines the desired state of ContentLibraryProvider
+type ContentLibraryProviderSpec struct {
+	// Name describes the name of a content library. This is not necessarily the same as vCenter
+	// name of the content library.
+	Name string `json:"name,omitempty"`
```
aruneshpa, line 14: This field is meant to represent the vSphere content library name. What do we think about making this field optional since we only need the UUID to uniquely identify a vSphere content library.

janitha: Yeah, optional seems right since the lookup is done by UUID, not by Name. 

Also, how about `OverrideName` (or something like that) to be more clear that this is to simply use a different name for the CL rather than the vSphere set one.

aruneshpa: How about we call it `vSphereName` or `vSphereLibraryName` (since `vSphereCLName` has abbreviations and `vSphereContentLibraryName` is too long)

api/v1alpha1/contentlibraryprovider_types.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ContentLibraryProviderSpec defines the desired state of ContentLibraryProvider
+type ContentLibraryProviderSpec struct {
+	// vSphereName describes the vSphere name of a content library provider. This is the name
+	// of the content library in the vSphere inventory view.
+	// +optional
+	VsphereName string `json:"vSphereName,omitempty"`
```
janitha, line 15: `VSphereName`?

api/v1alpha1/contentlibraryprovider_types.go
aruneshpa: Removed it since it was getting confusing to have two names in the kubectl output.

api/v1alpha1/contentlibraryprovider_types.go
aruneshpa: NA any more.

PR #16, opened 2020-05-28
Introduce OvfEnv VmMetadataTransport

janitha:
OvfEnv is a standard way to transport key/value vm customization data from the infrastructure to the VM, as well as exposing them to guest os/apps within the VM. Possible key/value properties that are userConfigurable would be already defined in the .ovf image as defined in the OVF Spec 2.1.1 Section 11 https://www.dmtf.org/sites/default/files/standards/documents/DSP0243_2.1.1.pdf 

This MR makes it possible to supply key/value fields to set the userConfigurable property fields on a .ovf during the VirtualMachine create time. The VirtualMachine spec already has a VmMetadata field which already supports a "ExtraConfig" transport, and this change adds another possible transport which is "OvfEnv".

### DIFF
--- a/api/v1alpha1/virtualmachine_types.go
+++ b/api/v1alpha1/virtualmachine_types.go
@@ -64,7 +64,7 @@ const (
 	VirtualMachineMetadataExtraConfigTransport VirtualMachineMetadataTransport = "ExtraConfig"
 
 	// VirtualMachineMetadataOvfEnvTransport will set the VirtualMachineMetadata ConfigMap data as
-	// vApp properties on the VM, which will be exposed as OvfEnv to the VM Guest. Only properties
+	// vApp properties on the VM, which will be exposed as OvfEnv to the Guest VM. Only properties
 	// marked userConfigurable and already present in either OVF Properties of a VirtualMachineImage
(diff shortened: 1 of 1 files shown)


api/v1alpha1/virtualmachine_types.go
```diff
@@ -55,6 +55,21 @@ type VirtualMachineNetworkInterface struct {
...
+// VirtualMachineMetadataTransport is used to indicate the transport used by VirtualMachineMetadata
+type VirtualMachineMetadataTransport string
+
+const (
+	// VirtualMachineMetadataExtraConfigTransport will set the VirtualMachineMetadata ConfigMap data as
+	// extraConfig key value fields on the VM.
+	VirtualMachineMetadataExtraConfigTransport VirtualMachineMetadataTransport = "ExtraConfig"
+
+	// VirtualMachineMetadataOvfEnvTransport will set the VirtualMachineMetadata ConfigMap data as
+	// vApp properties on the VM, which will be exposed as OvfEnv to the VM Guest. Only properties
```
aruneshpa, line 67: `VM Guest` -> `Guest VM`?

janitha: Fixed

PR #20, opened 2020-06-22
Add Conditions to VirtualMachine status

jainvarun21:
To report VM status, we currently use Phase and PowerState. Phase is a high level abstraction of VM's execution phase, and PowerState, as implied by its name conveys VM power-state. We currently do not have a means to convey any fine-grained status that is actionable from user/client's perspective. We do have Events, but they are generally meant to be used as an equivalent of log messages for human consumption and not intended for event-driven automation.

It would be useful to have an equivalent of Conditions array present in the core Kubernetes resources like Pods and Nodes to report fine-grained status information that is actionable from the client's perspective.


api/v1alpha1/virtualmachine_types.go
```diff
@@ -211,6 +211,9 @@ type VirtualMachineStatus struct {
 	// Phase describes the current phase information of the VirtualMachine.
 	Phase VMStatusPhase `json:"phase"`
 
+	// Conditions describes the current condition information of the VirtualMachine.
+	Conditions []VirtualMachineCondition `json:"conditions,omitempty"`
```
janitha, line 215: PodStatus and NodeStatus both does `patchStrategy:"merge" patchMergeKey:"type"` wondering if we need it here as well?

api/v1alpha1/virtualmachine_types.go
```diff
@@ -228,6 +231,27 @@ type VirtualMachineStatus struct {
...
+// VirtualMachineCondition contains condition information for a VirtualMachine.
+type VirtualMachineCondition struct {
+	// Type of the condition.
+	Type VirtualMachineConditionType `json:"type"`
+
+	// Status of the condition, one of ('True', 'False', 'Unknown').
+	Status metav1.ConditionStatus `json:"status"`
+
+	// The last time the condition transitioned from one status to another.
+	LastTransitionTime *metav1.Time `json:"lastTransitionTime,omitempty"`
```
janitha, line 243: Nit, pointer may not be necessary here.

PR #21, opened 2020-07-20
Add VirtualMachineDisk, SystemDisk to spec

dramdass:
Currently, only VirtualMachineVolumes can be specified in the
VM service. These are limited to CNS volumes and are to be
use by PVCSI addons in TKG clusters.

This change introduces disk support to the VM service. A
barebones VirtualMachineDisk struct is added and can be specified
as a SystemDisk for a VirtualMachine. Future commits may
introduce AttachedDisks to the VirtualMachineSpec to specify
desired attachment for non-system disks.


api/v1alpha1/virtualmachine_types.go
```diff
@@ -132,6 +132,12 @@ type TCPSocketAction struct {
 	Host string `json:"host,omitempty"`
 }
 
+// VirtualMachineDisk defines the properties of a disk.
+type VirtualMachineDisk struct {
+	// Size specifies the disk size in GB
+	Size int `json:"size"`
```
janitha, line 138: The size type should confirm to k8s convention https://github.com/kubernetes/community/blob/master/contributors/design-proposals/scheduling/resources.md

api/v1alpha1/virtualmachine_types.go
```diff
@@ -183,6 +189,10 @@ type VirtualMachineSpec struct {
 	// responding to the probe.
 	// +optional
 	ReadinessProbe *Probe `json:"readinessProbe,omitempty"`
+
+	// SystemDisk specifies the VirtualMachineDisk which is the SystemDisk of a VirtualMachine.
+	// +optional
+	SystemDisk *VirtualMachineDisk `json:"systemDisk,omitempty"`
```
janitha, line 195: For a given source vm template (eg. ovf image or existing inventory vm), if there are multiple disks, vsphere actually wouldn't know which one is the system disk. Will need something like the disklocator field to specify the disk by id (or something similar)

api/v1alpha1/virtualmachine_types.go
dramdass: Will fix. Thanks

api/v1alpha1/virtualmachine_types.go
dramdass: That's probably needed yeah. For the Gardener use case, it looks like it just assumes the first disk: https://github.com/gardener/machine-controller-manager-provider-vsphere/blob/master/pkg/vsphere/internal/clone.go#L565

We probably need to sure up this logic here and use something robust to identify

jainvarun21: If the plan is to include the ability to resize any/all disks in the future, there is value in considering modeling this field after deviceChange in ConfigSpec: https://pubs.vmware.com/vi3/sdk/ReferenceGuide/vim.vm.ConfigSpec.html.

dramdass: Thanks @jainvarun21 - for reference, the wiki page discussing this internally is at https://confluence.eng.vmware.com/display/WCP/Virtual+Disk+Specification

The plan is to be able to resize any/all disks in the future. It makes sense to refer to the VM config spec for modeling this field but I would like us to slowly adopt only the config we need right now since use cases may be better suited for simpler/modified config

dramdass:
Abandoning since latest design doc has different API schema. Will send new MR

PR #22, opened 2020-08-06
Add VsphereVolumeSource to API

dramdass:
The VsphereVolumeSource struct represents static disks belonging
to a VirtualMachine. This struct is specified in the
VirtualMachineSpec via embedding in the VirtualMachineVolume
struct.


dramdass:
Abandoning since we're using forking model

PR #23, opened 2020-08-06
Add VsphereVolumeSource to API

dramdass:
The VsphereVolumeSource struct represents static disks belonging
to a VirtualMachine. This struct is specified in the
VirtualMachineSpec via embedding in the VirtualMachineVolume
struct.


.gitignore
```diff
@@ -23,3 +23,6 @@ vendor
 *.swp
 *.swo
 *~
+
+$ generated config files
```
janitha, line 27: Seems like this should be a comment with a `#`

api/v1alpha1/virtualmachine_types.go
```diff
@@ -102,9 +102,27 @@ type VirtualMachineVolume struct {
 	// have a unique name.
 	Name string `json:"name"`
 
+	// StoragePolicy associated with volume.
+	StoragePolicy   *string `json:"storagePolicy,omitempty"`
```
janitha, line 106: StoragePolicy will only apply to some types of volumes. For example, it won't apply to PersistentVolumeClaims (if there is a storage policy, it would be defined with the PV itself). 

This field probably can be moved in to the VSphereVolumeSource struct. Otherwise, please expand the documentation to state the limitations.

api/v1alpha1/virtualmachine_types.go
```diff
@@ -102,9 +102,27 @@ type VirtualMachineVolume struct {
...
+
+// VsphereVolumeSource describes a volume source that represent static disks that belong to a VirtualMachine.
+type VsphereVolumeSource struct {
+	// A description of the virtual volume's resources and capacity
+	// +optional
+	Capacity        corev1.ResourceList `json:"capacity,omitempty"`
+
+	// Device key of vSphere disk.  Empty deviceKey means it should be created in vSphere.
+	// +optional
+	DeviceKey          *string `json:"deviceKey,omitempty"`
```
janitha, line 125: This can be of type int since vSphere device keys are ints

.gitignore
dramdass: Thanks for catching this typo!

api/v1alpha1/virtualmachine_types.go
dramdass: Thanks

api/v1alpha1/virtualmachine_types.go
dramdass: Per Derek's comment in the design doc, we should specify this at the VirtualMachineVolume level. As you rightly said, for PersistentVolumeClaims (if there is a storage policy, it would be defined with the PV itself). I'll update my comment to say it only current applies to the VSphereVolumeSource type. I think the reasoning behind this was to avoid duplicating that field in all the different volume types. We need to decide the override behavior for PVC however. I am assuming we don't want this to override the one in the PV spec

api/v1alpha1/virtualmachine_types.go
```diff
@@ -102,9 +102,40 @@ type VirtualMachineVolume struct {
...
+	// PersistentVolume spec.
+	// +optional
+	StoragePolicyName   *string `json:"storagePolicyName,omitempty"`
+
 	// PersistentVolumeClaim represents a reference to a PersistentVolumeClaim in the same namespace. The PersistentVolumeClaim
 	// must match a persistent volume provisioned (either statically or dynamically) by the cluster's CSI provider.
+	// +optional
 	PersistentVolumeClaim *corev1.PersistentVolumeClaimVolumeSource `json:"persistentVolumeClaim,omitempty"`
+
+	// VSphereVolume represents a reference to a vSphereVolumeSpec in the same namespace. Only one of PersistentVolumeClaim or
```
janitha, line 124: - Comment should be `VsphereVolumeSource` instead of `vSphereVolumeSpec`
- The mutual exclusive nature of `PersistentVolumeClaim` or `VSphereVolume` should be commented as part of the `VirtualMachineVolume struct` comment.

dramdass: Done!

PR #25, opened 2020-08-25
Introduce VirtualMachineClassBinding CRD

sreyasn:
The CRD represents a binding between a supervisor namespace and a VM class.
Similar to a VM class resource today, this will be a read-only resource for devops users.

For example,
if a supervisor namespace has 3 VirtualMachineClassBinding resources with classRefs pointing to 3 VirtualMachineClasses,
then that namespace has access to three VM classes (small, medium, large).


api/v1alpha1/virtualmachineclassbinding_types.go
```diff
@@ -0,0 +1,44 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ClassReference contains info to locate a specific VirtualMachineClass object
+type ClassReference struct {
+	// API version of the referent.
+	APIVersion string `json:"apiVersion,omitempty"`
+	// Kind is the type of resource being referenced.
+	Kind string `json:"kind"`
+	// Name is the name of resource being referenced.
```
janitha, line 16: What's the value in specifying the `APIVersion` and `Kind`? The referenced object will always be a VirtualMachiineClass, and a name is unique enough (it will be unique across APIVersions as well). 

I think just `Name` here is sufficient?

sreyasn: Yeah. I think `Kind` is redundant atm. APIVersion with an omitempty and Name should suffice.

api/v1alpha1/virtualmachineclassbinding_types.go
```diff
@@ -0,0 +1,40 @@
...
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ClassReference contains info to locate a Kind VirtualMachineClass object
+type ClassReference struct {
+	// Name is the name of resource being referenced.
```
aruneshpa, line 12: I think we should add the `ApiVersion` or `ApiGroup` and `Kind` since we want to make the association to a particular VirtualMachineClass explicit. `Kind` is sort of redundant, but I would rather have the association explicit than have a webhook default to an older version once we upgrade to a newer version.

sreyasn: I've made them available.

PR #26, opened 2020-09-18
Add default volume provisioning type for VirtualMachine

dramdass:
This change introduces additional configuration for VirtualMachine
that will allow users to specify what type of provisioning (e.g
thin or eager zeroed) for all the disks belonging to this
VirtualMachine. In a subsequent change, this configuration shall
also be available for each VirtualMachineVolume.

As part of this change, this configuration is added under a new
AdvancedOptions section of the VirtualMachine spec which was
planning to be introduced soon for such optional, advanced
configurations.

Testing done:
- make all


api/v1alpha1/virtualmachine_types.go
```diff
@@ -9,6 +9,22 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+type VolumeProvisioningType string
+
+// See OvfCreateImportSpecParamsDiskProvisioningType
+const (
+	VolumeProvisioningTypeMonolithicSparse     = VolumeProvisioningType("monolithicSparse")
```
janitha, line 16: Lot of these would be dependent on the datastore for support, and not applicable for most normal or advanced workflow (outside of the thick/thin and eager zero). 

How about introducing two new booleans, optional `thickProvisioned` (false to imply thin), and `eagerZeroed` (applicable only when thickProvisioned=true). Internally in vm-operator we can take those two booleans and pick the correct provisioning type string.

dramdass: I'm good with that. I'll make that change

dramdass: Looking at https://opengrok.eng.vmware.com/source/xref/main.perforce.1666/bora/vim/vmodl/vim/VirtualDiskManager.java#168 it makes more sense for me to have `thinProvisioned` instead of `thickProvisioned` be the bool. Any objections?

janitha: No objects, (alternatively we can use the terminology `preallocated` as well), no strong preference from me either way.

dramdass: Please re-open if you disagree. Thanks!

api/v1alpha1/virtualmachine_types.go
```diff
@@ -215,6 +215,30 @@ type VirtualMachineSpec struct {
...
 	ReadinessProbe *Probe `json:"readinessProbe,omitempty"`
+
+	// AdvancedOptions describes a set of optional, advanced options for configuring a VirtualMachine
+	AdvancedOptions *VirtualMachineAdvancedOptions `json:"advancedOptions,omitempty"`
+}
+
+// AdvancedOptions describes a set of optional, advanced options for configuring a VirtualMachine
+type VirtualMachineAdvancedOptions struct {
+	// DefaultProvisioningOptions specifies the provisioning type to be used by default for all VirtualMachine volumes.
+	// This can be overridden for each VirtualMachineVolume.
```
janitha, line 226: It's not possible to override it today, so would leave it this.

api/v1alpha1/virtualmachine_types.go
```diff
@@ -215,6 +215,30 @@ type VirtualMachineSpec struct {
...
 	// +optional
 	ReadinessProbe *Probe `json:"readinessProbe,omitempty"`
+
+	// AdvancedOptions describes a set of optional, advanced options for configuring a VirtualMachine
+	AdvancedOptions *VirtualMachineAdvancedOptions `json:"advancedOptions,omitempty"`
+}
+
+// AdvancedOptions describes a set of optional, advanced options for configuring a VirtualMachine
+type VirtualMachineAdvancedOptions struct {
+	// DefaultProvisioningOptions specifies the provisioning type to be used by default for all VirtualMachine volumes.
```
janitha, line 225: Clarification, only volumes (ie. disks) that are exclusively owned by this VM. 
Does not apply to PersistentVolumeClaim volumes that are created and managed externally.

janitha (approved):

PR #27, opened 2020-09-22
Introduce ChangeBlockTracking flag for Virtual Machines

janitha:
This change introduces a flag to enable Change Block Tracking as an
optional boolean as part of the virtual machine advanced opotions.

Testing done: `make all`


api/v1alpha1/virtualmachine_types.go
```diff
@@ -225,6 +225,9 @@ type VirtualMachineAdvancedOptions struct {
 	// DefaultProvisioningOptions specifies the provisioning type to be used by default for VirtualMachine volumes exclusively
 	// owned by this VirtualMachine. This does not apply to PersistentVolumeClaim volumes that are created and managed externally.
 	DefaultVolumeProvisioningOptions *VirtualMachineVolumeProvisioningOptions `json:"defaultVolumeProvisioningOptions,omitempty"`
+
+	// ChangeBlockTracking specifies the enablement of change block tracking on the VirtualMachine.
```
dramdass, line 229: Is there a standard VMW one liner for summarizing what CBT is? And would be it the right place to add it here?

dramdass (approved):
LGTM - 1 minor question/comment

api/v1alpha1/virtualmachine_types.go
janitha: Good idea, updated with a short summary stating it's meant for incremental backups.

PR #28, opened 2020-09-22
Add ProviderRef to VirtualMachineNetworkInterface

sreyasn:
The NetworkInterfaceProviderRef will allow a user to specify the
VM Network interface config through a provider (a Netoperator Network
Interface object) for VMoperator to consume.


dramdass:
LGTM!

dramdass (approved):

api/v1alpha1/virtualmachine_types.go
```diff
@@ -47,6 +47,18 @@ type VirtualMachinePort struct {
...
+// NetworkInterfaceProviderReference contains info to locate a network interface provider object.
+type NetworkInterfaceProviderReference struct {
+	// APIGroup is the group for the resource being referenced.
+	APIGroup string `json:"apiGroup"`
+	// Kind is the type of resource being referenced
+	Kind string `json:"kind"`
+	// Name is the name of resource being referenced
+	Name string `json:"name"`
+	// API version of the referent.
+	APIVersion string `json:"apiVersion,omitempty"`
```
janitha, line 59: Why omitempty just for this?

janitha (approved):

api/v1alpha1/virtualmachine_types.go
```diff
@@ -61,6 +73,10 @@ type VirtualMachineNetworkInterface struct {
 	// NSX-T networks, this is the name of a pre-existing NSX-T VirtualNetwork.
 	NetworkName string `json:"networkName"`
 
+	// ProviderRef is reference to a network interface provider object that specifies the network interface configuration.
+	// If unset, default configuration is assumed.
+	ProviderRef NetworkInterfaceProviderReference `json:"providerRef"`
```
bryanv, line 78: I would have expected this to be a pointer because it is optional.

api/v1alpha1/virtualmachine_types.go
sreyasn: Good question. I was using a pattern similar to our netop types here. 

But seems like the use case would be when we have a fallback version later down the line. v1a1 or v1.

PR #29, opened 2020-10-05
Introduce changeBlockTracking status flag

janitha:
The changeBlockTracking status flag describes if CBT has been
explicitly enabled or disabled on the VM.

This is also necessary for the implementation of setting the
changeBlockTracking in spec, since the enablement is a multi step
process and this flag would enable setting in a deterministic way.


dramdass (approved):
LGTM!

sreyasn (approved):
LGTM @janitha .

PR #30, opened 2020-10-19
Introduce ContentSourceBinding CRD

jainvarun21:
This patch introduces ContentSourceBinding CRD that binds a
ContentSource to a supervisor namespace. An instance of this CRD
in a supervisor namespace indicates that the namespace has access
to the ContentSource it points to.


api/v1alpha1/contentsourcebinding_types.go
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ContentSourceReference contains info to locate a Kind ContentSource object
```
jainvarun21, line 10: Wonder if we should club this with ClassReference (https://github.com/vmware-tanzu/vm-operator-api/blob/master/api/v1alpha1/virtualmachineclassbinding_types.go#L11) into a common ObjectReference.

Didn't venture into it in this PR since that would be a breaking api change.

janitha: The k8s docs recommend using a focused class such as yours instead of a using ObjectReference https://godoc.org/k8s.io/api/core/v1#ObjectReference so keeping it as is seems okay

> Instead of using this type, create a locally provided and used type that is well-focused on your reference. For example, ServiceReferences for admission registration: https://github.com/kubernetes/api/blob/release-1.17/admissionregistration/v1/types.go#L533 . +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object

api/v1alpha1/contentsourcebinding_types.go
```diff
@@ -0,0 +1,45 @@
...
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Namespaced,shortName=contentsourcebinding
+// +kubebuilder:printcolumn:name="ContentSource",type="string",JSONPath=".contentSourceRef.name"
+
+// ContentSourceBinding is a binding object responsible for
+// defining a ContentSource and a Namespace associated with it
+type ContentSourceBinding struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
```
janitha, line 29: Since ContentSourceBinding is a root object, it should have a Spec (and optionally a Status), see other CRDs.

api/v1alpha1/contentsourcebinding_types.go
```diff
@@ -0,0 +1,45 @@
...
+	// ContentSourceRef is a reference to a ContentSource object
+	ContentSourceRef ContentSourceReference `json:"classRef,omitempty"`
```
janitha, line 31: The json annotation should be corrected to `contentSourceRef`

api/v1alpha1/contentsourcebinding_types.go
jainvarun21: I see. Thanks for the information, Janitha.

api/v1alpha1/contentsourcebinding_types.go
janitha: Resolving as per offline discussion, I don't feel strongly either this or using a spec/status

janitha (approved):

api/v1alpha1/contentsourcebinding_types.go
aruneshpa: Yes. Using an `ObjectReference` is not preferred.

api/v1alpha1/contentsourcebinding_types.go
```diff
@@ -0,0 +1,45 @@
...
+	// API version of the referent.
+	APIVersion string `json:"apiVersion,omitempty"`
+	// Kind is the type of resource being referenced.
+	Kind string `json:"kind,omitempty"`
+	// Name is the name of resource being referenced.
+	Name string `json:"name"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Namespaced,shortName=contentsourcebinding
```
aruneshpa, line 21: Having the `shortName` as the same name as the resource doesnt add a lot of value. Maybe we should name it to `csbinding` or even `csb`?  Historically, `shortname` is supposed to be really short (E.g. `ig`, `ma` etc. 

https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/markers/crd.go#L258)

api/v1alpha1/contentsourcebinding_types.go
aruneshpa, line 10: Here and below: can you make sure the type comments end with a period(`.`)?

api/v1alpha1/contentsourcebinding_types.go
```diff
@@ -0,0 +1,45 @@
...
+// +kubebuilder:printcolumn:name="ContentSource",type="string",JSONPath=".contentSourceRef.name"
+
+// ContentSourceBinding is a binding object responsible for
+// defining a ContentSource and a Namespace associated with it
```
aruneshpa, line 25: To some extent, this comment reads that this object has anything to do with `ContentSource` definition. 

Maybe rename it as 
```ContentSourceBinding is an object that represents a ContentSource to Namespace mapping.```

File: api/v1alpha1/contentsourcebinding_types.go [hunk 6]
```diff
@@ -0,0 +1,45 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// ContentSourceRef is a reference to a ContentSource object
+	ContentSourceRef ContentSourceReference `json:"contentSourceRef,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+
+// ContentSourceBindingList contains a list of ContentSourceBinding
```
aruneshpa, line 36: `ContentSourceBinding` -> `ContentSourceBindings`?

aruneshpa (approved):
LGTM with minor nits.

api/v1alpha1/contentsourcebinding_types.go
bryanv: I think we should omit the shortname, for now. We don't have it one the corresponding ContentSource.

api/v1alpha1/contentsourcebinding_types.go
bryanv: The format of this string is how kubebuilder would have generated it, and we've kept it that way (non-plural) in the other files here.

api/v1alpha1/contentsourcebinding_types.go
jainvarun21: I have omitted the shortName in the latest diff. Thanks.

api/v1alpha1/contentsourcebinding_types.go
jainvarun21: Agree. I have changed the comment based on your suggestion.

api/v1alpha1/contentsourcebinding_types.go
jainvarun21: Resolving this conversation as per Bryan's comment.

api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -188,6 +203,64 @@ func (in *ContentSource) DeepCopyObject() runtime.Object {
 	return nil
 }
 
+// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
+func (in *ContentSourceBinding) DeepCopyInto(out *ContentSourceBinding) {
+	*out = *in
+	out.TypeMeta = in.TypeMeta
+	in.ObjectMeta.DeepCopyInto(&out.ObjectMeta)
+	out.ClassRef = in.ClassRef
```
sreyasn, line 211: References to a ClassRef in ContentSourceBinding?. seems like stale code

jainvarun21: Correct. Thanks for catching this. I'll send an MR to correct this.

jainvarun21: Fixed it here: https://github.com/vmware-tanzu/vm-operator-api/pull/32/

PR #31, opened 2020-10-21
Add SupportedOS boolean flag to the VirtualMachineImage status 

sreyasn:
InvalidGuestOSType is used to indicate that the GuestOS Type in VirtualMachineImageOSInfo is Invalid
and helps with validation during  VirtualMachineImage creation

### DIFF
--- a/api/v1alpha1/virtualmachineimage_types.go
+++ b/api/v1alpha1/virtualmachineimage_types.go
@@ -64,9 +64,10 @@ type VirtualMachineImageSpec struct {
 
 // VirtualMachineImageStatus defines the observed state of VirtualMachineImage
 type VirtualMachineImageStatus struct {
-	Uuid       string `json:"uuid,omitempty"`
-	InternalId string `json:"internalId"`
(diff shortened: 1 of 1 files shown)


dramdass:
LGTM

dramdass (approved):

janitha (commented):
Instead of overloading the VirtualMachineImageOSInfo.Type to indicate if the os is valid or not, perhaps we should make the validity state of a virtual machine part of the status, thoughts? (can discuss in the correspinding vm-operator MR as well)

aruneshpa:
LGTM

bryanv:
> Instead of overloading the VirtualMachineImageOSInfo.Type to indicate if the os is valid or not, perhaps we should make the validity state of a virtual machine part of the status, thoughts? (can discuss in the correspinding vm-operator MR as well)

I too have the same thought. The Type is the Type, and the validity - or perhaps "supported" - is really determined by the current VC.

janitha (approved):
Glad we can utilize status for this purpose

PR #33, opened 2020-10-22
Update VM Probe configuration

yi0909:
This adds PeriodSeconds to probe configuration, which indicates how often to perform the probe.
It also adds VirtualMachineReady condition type in virtualmachine status.

### DIFF
--- a/api/v1alpha1/virtualmachine_types.go
+++ b/api/v1alpha1/virtualmachine_types.go
@@ -165,7 +165,17 @@ type Probe struct {
 	// TimeoutSeconds specifies a number of seconds after which the probe times out.
 	// Defaults to 10 seconds. Minimum value is 1.
 	// +optional
+	// +kubebuilder:default:=10
+	// +kubebuilder:validation:Minimum:=1
+	// +kubebuilder:validation:Maximum:=60
 	TimeoutSeconds int32 `json:"timeoutSeconds,omitempty"`
+
+	// PeriodSeconds specifics how often (in seconds) to perform the probe.
+	// Defaults to 10 seconds. Minimum value is 1.
+	// +optional
+	// +kubebuilder:default:=10
+	// +kubebuilder:validation:Minimum:=1
+	PeriodSeconds int32 `json:"periodSeconds,omitempty"`
 }
 
 // TCPSocketAction describes an action based on opening a socket.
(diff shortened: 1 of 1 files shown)


janitha:
Please run `make generate` and include the zz_generated.deepcopy.go in the PR please

api/v1alpha1/virtualmachine_types.go
```diff
@@ -166,6 +167,11 @@ type Probe struct {
 	// Defaults to 10 seconds. Minimum value is 1.
 	// +optional
 	TimeoutSeconds int32 `json:"timeoutSeconds,omitempty"`
+
+	// PeriodSeconds specifics how often (in seconds) to perform the probe.
+	// Defaults to 10 seconds. Minimum value is 1.
+	// +optional
+	PeriodSeconds int32 `json:"periodSeconds,omitempty"`
```
janitha, line 174: Could you sprinkle in some kubebuilder validation markers for default and minimum here and for TimeoutSeconds above. See https://book.kubebuilder.io/reference/markers/crd-validation.html

api/v1alpha1/virtualmachine_types.go
```diff
@@ -13,8 +13,9 @@ type VirtualMachinePowerState string
 
 // See govmomi.vim25.types.VirtualMachinePowerState
 const (
-	VirtualMachinePoweredOff VirtualMachinePowerState = "poweredOff"
-	VirtualMachinePoweredOn  VirtualMachinePowerState = "poweredOn"
+	VirtualMachinePoweredOff VirtualMachinePowerState    = "poweredOff"
+	VirtualMachinePoweredOn  VirtualMachinePowerState    = "poweredOn"
+	VirtualMachineReady      VirtualMachineConditionType = "Ready"
```
bryanv, line 18: This doesn't belong here. Start a new const for the VirtualMachineConditionType's, that's below the type definition like is done elsewhere in this file.

api/v1alpha1/virtualmachine_types.go
```diff
@@ -166,6 +167,11 @@ type Probe struct {
 	// Defaults to 10 seconds. Minimum value is 1.
 	// +optional
 	TimeoutSeconds int32 `json:"timeoutSeconds,omitempty"`
```
bryanv, line 169: Part implementation detail, and also that these probes are not execute via anything similar to kubelet, I think we need to either define a maximum here, or internally cap the provided timeout.

bryanv:
> Please run `make generate` and include the zz_generated.deepcopy.go in the PR please

I would not expect the generated Probe DeepCopy() to change from this change. The new field is just an `int32` and copied for free because of the struct assignment.

api/v1alpha1/virtualmachine_types.go
yi0909: Yeah we can define a maximum here. Is 30 seconds a reasonable value for the maximum timeout?

janitha (approved):

PR #35, opened 2020-10-26
Make SupportedOS GOSC specific

sreyasn:
Taking into consideration that GOSC could become
optional and SupportedOS could be misleading for
failures specific to GuestOS customization.

Also making the field display on higher priority
/w '-o wide'


api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -64,10 +64,10 @@ type VirtualMachineImageSpec struct {
...
 // VirtualMachineImageStatus defines the observed state of VirtualMachineImage
 type VirtualMachineImageStatus struct {
-	Uuid        string `json:"uuid,omitempty"`
-	InternalId  string `json:"internalId"`
-	PowerState  string `json:"powerState,omitempty"`
-	SupportedOS *bool  `json:"supportedOS,omitempty"`
+	Uuid          string `json:"uuid,omitempty"`
+	InternalId    string `json:"internalId"`
+	PowerState    string `json:"powerState,omitempty"`
+	GOSCSupported *bool  `json:"goscSupported,omitempty"`
```
janitha, line 70: Could you add some comments for the fields here inside VirtualMaciineImageStatus (the other ones three are pretty self explanatory but lets add them for completeness), mainly GOSCSupported will definitely need some description since it may not be clear to a lot of people.

bryanv (approved):

PR #37, opened 2020-10-26
Remove default value for TimeoutSeconds and PeriodSeconds.

yi0909:
We use API version v1beta1, which does not support defaulting.


sreyasn (approved):
LGTM. thanks for the fix

PR #38, opened 2020-10-27
Rename GOSC_Supported to GuestOSCustomizable  and Update description for the VirtualMachineImage Status

sreyasn:
Rename GOSC_Supported to make it clear for the end user and also

Update description for the VirtualMachineImage Status fields (ie)
Mark UUID, internalID and PowerState fields as deprecated


api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -64,8 +64,14 @@ type VirtualMachineImageSpec struct {
 
 // VirtualMachineImageStatus defines the observed state of VirtualMachineImage
 type VirtualMachineImageStatus struct {
+	// Uuid uniquely identifies the VirtualMachineImage instance on Virtual Center
```
janitha, line 67: The Uuid and InternalId are dependent on the source of the virtual machine image. 

For CL images, the Uuid is the content library item's ID, and internalId is the name (not sure why we have it that way, since the object's metadata.name is also the same). 

For inventory VM images, Uuid is the vCenter's VM's BiosUUID (ugh, we need to stop using that since it's not unique across vms), and InternalId is the moRef. 

We really need to do redo the status field, I've added it as an item for v1alpha2.

sreyasn: Makes sense. a bunch of things to rework there. Let's not worry about fixing the comments for this now?.

sreyasn: Do we just wanna add comments on top saying // Deprecated ?

janitha: Let's do deprecated

api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -64,13 +64,18 @@ type VirtualMachineImageSpec struct {
...
+
+	// Deprecated
 	PowerState    string `json:"powerState,omitempty"`
 
-	// GOSCSupported describes whether the VirtualMachineImage is GuestOS customizable based on
-	// supported OSTypes for the cluster
-	GOSCSupported *bool  `json:"goscSupported,omitempty"`
+	// GuestOSCustomizable indicates whether the VirtualMachineImage's osType is supported for
+	// customisation on the cluster
+	GuestOSCustomizable *bool  `json:"goscSupported,omitempty"`
```
janitha, line 78: The json marker needs to be updated as well

sreyasn: ah! Good catch

janitha (approved):

PR #39, opened 2020-10-30
Add loadBalancerIP field in VirtualMachineService

maplain:
1. Being able to statically assign IP addresses to LoadBalancer type VirtualMachineService is a very common ask from users;
2. Adding this field enables this functionality in the API implementation/Providers

### DIFF
--- a/api/v1alpha1/virtualmachineservice_types.go
+++ b/api/v1alpha1/virtualmachineservice_types.go
@@ -85,6 +85,14 @@ type VirtualMachineServiceSpec struct {
 	// VirtualMachineService with the set of VirtualMachines that should back this VirtualMachineService.
 	Selector map[string]string `json:"selector"`
 
+	// Only applies to Service Type: LoadBalancer
+	// LoadBalancer will get created with the IP specified in this field.
+	// This feature depends on whether the underlying loadbalancer provider supports specifying
+	// the loadBalancerIP when a load balancer is created.
+	// This field will be ignored if the provider does not support the feature.
+	// +optional
+	LoadBalancerIP string `json:"loadBalancerIP,omitempty"`
+
 	// Just support cluster IP for now
 	ClusterIP    string `json:"clusterIp,omitempty"`
 	ExternalName string `json:"externalName,omitempty"`


maplain:
@corrieb @sreyasn could you please take a look?

janitha:
The `LoadBalancerIP` will be mutually exclusive with the `ClusterIP` right? If so, can you also update the comments for ClusterIP as well stating so.

maplain:
@janitha thx for your response!

no, I don't think they're mutually exclusive. loadBalancerIP is used to specify the static IP for external LoadBalancer when Service is type LoadBalancer and the LB provider supports it.
ClusterIP is used to specify the service's internal Cluster IP, implemented by kube-proxy.

Actually, this field is yet another field we copied from the upstream corev1.Serivce.spec: https://github.com/kubernetes/api/blob/master/core/v1/types.go#L4124

janitha (approved):

maplain:
@janitha thx a lot!

do you mind merging it? I don't seem to have the `write access`...

image

api/v1alpha1/virtualmachineservice_types.go
```diff
@@ -85,6 +85,14 @@ type VirtualMachineServiceSpec struct {
...
 
+	// Only applies to Service Type: LoadBalancer
+	// LoadBalancer will get created with the IP specified in this field.
+	// This feature depends on whether the underlying loadbalancer provider supports specifying
+	// the loadBalancerIP when a load balancer is created.
+	// This field will be ignored if the provider does not support the feature.
+	// +optional
+	LoadBalancerIP string `json:"loadBalancerIP,omitempty"`
+
 	// Just support cluster IP for now
```
bryanv, line 96: Can you remove this comment that would appear to not apply anymore? 

It is unfortunate that `clusterIp` instead of `clusterIP` was used in for the json name for ClusterIP.

maplain: yeah, sure: https://github.com/vmware-tanzu/vm-operator-api/pull/41

PR #41, opened 2020-11-09
Clean up comment

maplain:
1. Other types, eg: LoadBalancer is also supported besides Client IP


maplain:
@bryanv

janitha (approved):
Thanks for the change

PR #44, opened 2020-11-17
Add LoadBalancerSourceRanges to VirtualMachineService spec

maplain:
1. This PR adds one more field `LoadBalancerSourceRanges` to the VirtualMachineService spec.
   When supported by the underlying load balancer provider, ingress
   traffic will be restricted to the specified list of Client IPs

### DIFF
--- a/api/v1alpha1/virtualmachineservice_types.go
+++ b/api/v1alpha1/virtualmachineservice_types.go
@@ -93,6 +93,14 @@ type VirtualMachineServiceSpec struct {
 	// +optional
 	LoadBalancerIP string `json:"loadBalancerIP,omitempty"`
 
+	// LoadBalancerSourceRanges is an array of IP addresses in the format of
+	// CIDRs, for example: 103.21.244.0/22 and 10.0.0.0/24.
+	// If specified and supported by the platform, this will restrict
+	// ingress traffic to the specified client IPs. This field will be ignored if the
+	// loadbalancer provider does not support the feature.
+	// +optional
+	LoadBalancerSourceRanges []string `json:"loadBalancerSourceRanges,omitempty"`
+
 	ClusterIP    string `json:"clusterIp,omitempty"`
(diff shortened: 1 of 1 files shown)


maplain:
@janitha @bryanv @sreyasn

api/v1alpha1/virtualmachineservice_types.go
```diff
@@ -93,6 +93,13 @@ type VirtualMachineServiceSpec struct {
 	// +optional
 	LoadBalancerIP string `json:"loadBalancerIP,omitempty"`
 
+	// LoadBalancerSourceRanges is an array of IP addresses.
```
janitha, line 96: I would think this would also take network notation in addition to individual ip addresses right? (eg. 10.0.0.0/16 type notation?)

janitha (approved):

api/v1alpha1/virtualmachineservice_types.go
maplain: Good point! I updated the comment to make it clear that the string should be in the format of CIDRs.

api/v1alpha1/virtualmachineservice_types.go
```diff
@@ -93,6 +93,30 @@ type VirtualMachineServiceSpec struct {
...
+	//   spec:
+	//     ports:
+	//     - port: 443
+	//       targetPort: 9376
+	//     selector:
+	//       app: example
+	//     type: LoadBalancer
+	//     loadBalancerSourceRanges:
+	//     - 103.21.244.0/22
+	//     - 103.22.200.0/22
```
janitha, line 113: Thanks for expanding the comment thoroughly :) but I don't think this level of detailed example is appropriate for a single field. I think simply stating something like "... in the format of CIDRs, for example 103.21.244.0/22 and 1.2.3.4" would be sufficient.

maplain: makes sense. updated.

PR #47, opened 2020-12-03
Add generic Conditions derived from cluster-api

bryanv:
This does replace VirtualMachineCondition with Condition structure.
We haven't started to set conditions yet, and this doesn't change
the REST API so I don't think it is really breaking, but we could
probably modify GetConditions/SetConditions to convert between the
types if desired.

The GetConditions/SetConditions methods are so the VirtualMachine
satisfies the special Patch helper Conditions support.


api/v1alpha1/condition_consts.go
```diff
@@ -0,0 +1,36 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
janitha, line 2: This should be VMware and consistent with the other files

```
// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
// SPDX-License-Identifier: Apache-2.0
```

api/v1alpha1/condition_types.go
```diff
@@ -0,0 +1,81 @@
...
+	Reason string `json:"reason,omitempty"`
+
+	// A human readable message indicating details about the transition.
+	// This field may be empty.
+	// +optional
+	Message string `json:"message,omitempty"`
+}
+
+// Conditions provide observations of the operational state of a VM Operator API resource.
+type Conditions []Condition
```
janitha, line 81: What's the benefit of having this type instead of using []Condition?

janitha (approved):

api/v1alpha1/condition_consts.go
bryanv: I guess changing this file is probably fine, but not for condition_types.go since it is basically straight from cluster-api

api/v1alpha1/condition_types.go
bryanv: Maybe little, but the imported cluster-api code uses this type and I don't think it is worth increasing our changes to remove it

aruneshpa (approved):
LGTM.

PR #49, opened 2020-12-04
Make NetworkName an optional parameter

jainvarun21:
This patch provides the ability for clients to specify a
VirtualNetworkInterface (VIF) for a VM without specifying a
specific network by making NetworkName an optional parameter.
The network interface provider will then connect this VIF to
the default network for the namespace.

Testing Done:

- make all


dramdass (approved):
LGTM!

sreyasn (approved):
LGTM

PR #50, opened 2020-12-08
Update description for NetworkName

jainvarun21:
Update the description for NetworkName clarifying that
it is optional only if the NetworkType is "nsx-t".

Testing Done:

- make all

### DIFF
--- a/api/v1alpha1/virtualmachine_types.go
+++ b/api/v1alpha1/virtualmachine_types.go
@@ -70,9 +70,10 @@ type VirtualMachineNetworkInterface struct {
 	// +optional
 	NetworkType string `json:"networkType,omitempty"`
 
-	// NetworkName describes the name of an existing virtual network that this interface should be added to.  For
-	// NSX-T networks, this is the name of a pre-existing NSX-T VirtualNetwork. If unspecified, the network
-	// interface provider will connect this interface to the default network for the namespace.
+	// NetworkName describes the name of an existing virtual network that this interface should be added to.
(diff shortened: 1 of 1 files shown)


janitha (approved):

api/v1alpha1/virtualmachine_types.go
```diff
@@ -71,8 +71,9 @@ type VirtualMachineNetworkInterface struct {
 	NetworkType string `json:"networkType,omitempty"`
 
 	// NetworkName describes the name of an existing virtual network that this interface should be added to.  For
-	// NSX-T networks, this is the name of a pre-existing NSX-T VirtualNetwork. If unspecified, the network
-	// interface provider will connect this interface to the default network for the namespace.
+	// NSX-T networks, this is the name of a pre-existing NSX-T VirtualNetwork. If unspecified for "nsx-t" NetworkType,
```
bryanv, line 74: I suggest tweaking this a little for consistency:
   - In the first sentence, refer to '"nsx-t" NetworkType' instead of mentioning that in the second sentence. The second sentence could probably read "If unspecified, the default network for the namespace will be used."
   - For VDS, the sentence can start with the type, and read something like 'For "vsphere-distributed" NetworkType, the NetworkName must be specified.'

jainvarun21: Thanks for the review. Tweaked it as per your suggestion in the latest diff.

bryanv (approved):

PR #52, opened 2020-12-17
Add Condition and Reasons for VirtualMachineClassBindings and VirtualMachineClasses

aruneshpa:
There are many developer workflows where do not provide any user feedback. This
change defines a Condition and a few reasons to address some of those
workflows.

The following Condition is defined:
- VirtualMachinePreReqReady:
A condition that represents that all the Prerequisites of a `VirtualMachine` are
ready. For now, we only check for the availability of `VirtualMachineClass` and
the that the class is associated with the VirtualMachine's namespace. In future,
this Condition can be used to check for other components as well
(e.g. `VirtualMachineImage`, `StorageClass` etc).

A few reasons are also defined:
- VirtualMachineClassBindingNotFound
If the VirtualMachine's namespace does not have access to the specified
`VirtualMachineClass`.
- VirtualMachineClassNotFound
If the specified `VirtualMachineClass` does not exist on the cluster.

### DIFF
--- a/api/v1alpha1/condition_consts.go
+++ b/api/v1alpha1/condition_consts.go
@@ -9,15 +9,31 @@ const (
 	ReadyCondition ConditionType = "Ready"
 )
 
+// Conditions and condition Reasons for the VirtualMachine object.
+
+const (
+	// VirtualMachinePrereqReadyCondition documents that all of a VirtualMachine's prerequisites declared in the spec
+	// (e.g. VirtualMachineClass) are satisfied.
+	VirtualMachinePrereqReadyCondition ConditionType = "VirtualMachinePrereqReady"
+
+	// VirtualMachineClassBindingNotFoundReason (Severity=Error) documents a missing VirtualMachineClassBinding for the
+	// VirtualMachineClass specified in the VirtualMachineSpec.
+	VirtualMachineClassBindingNotFoundReason = "VirtualMachineClassBindingNotFound"
+
+	// VirtualMachineClassNotFoundReason (Severity=Error) documents that the VirtualMachineClass specified in the VirtualMachineSpec
+	// is not available.
+	VirtualMachineClassNotFoundReason = "VirtualMachineClassNotFound"
+)
+
 // Common Condition.Reason used by VM Operator API objects.
 const (
-	// DeletingReason (Severity=Info) documents an condition not in Status=True because the underlying object it is currently being deleted.
+	// DeletingReason (Severity=Info) documents a condition not in Status=True because the underlying object it is currently being deleted.
 	DeletingReason = "Deleting"
 
-	// DeletionFailedReason (Severity=Warning) documents an condition not in Status=True because the underlying object
+	// DeletionFailedReason (Severity=Warning) documents a condition not in Status=True because the underlying object
 	// encountered problems during deletion. This is a warning because the reconciler will retry deletion.
 	DeletionFailedReason = "DeletionFailed"
 
-	// DeletedReason (Severity=Info) documents an condition not in Status=True because the underlying object was deleted.
+	// DeletedReason (Severity=Info) documents a condition not in Status=True because the underlying object was deleted.
 	DeletedReason = "Deleted"
 )


janitha:
Prerequisite is a single word, so no need to uppercase the R in the middle

api/v1alpha1/condition_consts.go
```diff
@@ -9,15 +9,33 @@ const (
 	ReadyCondition ConditionType = "Ready"
 )
 
+const (
+	// VirtualMachinePreReqReadyCondition is set on VirtualMachines when all of a VirtualMachine's prerequisites are satisfied.
+	// For now, this check only validates that the VirtualMachineClass specified in the VirtualMachineSpec exists on the API
+	// server and the VirtualMachine's namespace has access to the specified VirtualMachineClass.
+	VirtualMachinePreReqReadyCondition = "VirtualMachinePreReqReady"
+
+	// VirtualMachinePreReqReadyReason documents that all of a VirtualMachine's PreRequisites have been satisfied.
```
janitha, line 18: Since the condition type already identifies that this is for VirtualMachinePreReqReady, seems redundant to also have the Reason be that verbose, it can just be "Ready" like other conditions.

janitha (approved):

api/v1alpha1/condition_consts.go
```diff
@@ -9,15 +9,33 @@ const (
 	ReadyCondition ConditionType = "Ready"
 )
 
+const (
+	// VirtualMachinePreReqReadyCondition is set on VirtualMachines when all of a VirtualMachine's prerequisites are satisfied.
+	// For now, this check only validates that the VirtualMachineClass specified in the VirtualMachineSpec exists on the API
```
bryanv, line 14: I don't think we need to call out "on the API server"; just exists is sufficient.

I'd reword this so we don't have to constantly update it, and for it to become out of date when we do. Say something along the lines of the reasons below are satisfied.

api/v1alpha1/condition_consts.go
```diff
@@ -9,15 +9,33 @@ const (
 	ReadyCondition ConditionType = "Ready"
 )
 
+const (
```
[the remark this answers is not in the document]
aruneshpa: Done.

jainvarun21:
Would this condition be relevant only during VM create? Have we considered an alternate workflow wherein we do not let the user create the VM k8s resource if the pre-reqs are not met (e.g. via webhooks)?

aruneshpa:
> Would this condition be relevant only during VM create?
@jainvarun21 , these conditions will be exposed in `Create` and `Update` VM paths. We want to expose these conditions as soon as possible so developers can take appropriate action (e.g. ask VI admins to add the VM class to a namespace).

>  Have we considered an alternate workflow wherein we do not let the user create the VM k8s resource if the pre-reqs are not met (e.g. via webhooks)?
Right now, we do enforce some of these constraints (checking that the image and class is specified). Since the VM class attach/detach is handles by wcpsvc asynchronously, it is perfectly fine for a VM to wait in the reconciler thread for the class binding to be available.

api/v1alpha1/condition_consts.go
```diff
@@ -9,15 +9,34 @@ const (
...
 )
 
+// Conditions and condition Reasons for the VirtualMachine object.
+
+const (
+	// VirtualMachinePrereqReadyCondition documents that all of a VirtualMachine's prerequisites declared in the spec
+	// (e.g. VirtualMachineClass) are satisfied.
+	VirtualMachinePrereqReadyCondition ConditionType = "VirtualMachinePrereqReady"
+
+	// VirtualMachinePrereqReadyReason documents that all of a VirtualMachine's PreRequisites have been satisfied.
```
jainvarun21, line 19: Do we need to set a Reason while the condition is true?

aruneshpa: No.

https://github.com/kubernetes-sigs/cluster-api/blob/dadd7bd6233edb1e422e7af282e923abff52f830/util/conditions/setter.go#L109

jainvarun21: In that case do we need VirtualMachinePrereqReadyReason?

aruneshpa: Removed the `Reason` for the True condition. Thanks for pointing that out.

PR #53, opened 2021-01-13
Change GuestOSCustomizable to SupportedGuestOS

sreyasn:
The GuestOSDescriptor ids fetched from the cluster
configoptions don't indicate whether the osType is customizable by GOSC or not.
It is a cautionary warning for GuestOS compatibility with the esx hosts of the
cluster.

It makes sense to change the name to SupportedGuestOS
since it will be used to mark osTypes that are considered "unsupported"
as false.


api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -73,9 +73,9 @@ type VirtualMachineImageStatus struct {
 	// Deprecated
 	PowerState    string `json:"powerState,omitempty"`
 
-	// GuestOSCustomizable indicates whether the VirtualMachineImage's osType is supported for
-	// customisation on the cluster
-	GuestOSCustomizable *bool  `json:"guestOsCustomizable,omitempty"`
+	// SupportedGuestOS indicates whether the VirtualMachineImage's osType is supported
+	// by the hosts in the cluster
```
janitha, line 77: Supported by hosts *and VM Service*

Because if it's a windows vm for example, the hosts will support it, but we (vm service) won't.

janitha (approved):

api/v1alpha1/virtualmachineimage_types.go
sreyasn: agreed. thanks.

api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -73,9 +73,9 @@ type VirtualMachineImageStatus struct {
...
+	SupportedGuestOS *bool  `json:"supportedGuestOs,omitempty"`
```
bryanv, line 78: I think this should be `supportedGuestOS`

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#constants 

>  When an acronym or initialism each letter in the acronym should be uppercase, such as with ClientIP or TCPDelay

bryanv (approved):

PR #55, opened 2021-01-27
VMSVC-513 Add imageCompatibility field to denote if an OVF is support…

srajashekar-vmw:
Due to the raciness between GOSC and cloud-init customization, we need a workflow that ensures VMs are created
successfully with customization using the VM Operator workflow.
This enforces VI admin to prepare the OVA using the instructions provided here https://confluence.eng.vmware.com/display/WCP/Pre-Requisite+Steps+to+Prepare+Ubuntu+OVA+for+VM+creation+using+VMSvc.
The preparation steps  also involves adding a new vApp option "vmware-system.vmsvc.compatibility" that denotes whether the image is being  prepared or not for VMService

This MR adds imageCompatibility field to VirtualMachineImageStatus API


api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -73,6 +73,10 @@ type VirtualMachineImageStatus struct {
 	// Deprecated
 	PowerState    string `json:"powerState,omitempty"`
 
+	// ImageCompatibility indicates whether the VirtualMachineImage's OVF is prepared by VI admin or not and is supported
```
aruneshpa, line 76: Do we only plan to set this for the OVF type images? Maybe we should reword this so this does not reference the image format?

aruneshpa (approved):

janitha:
A bit concerned about introducing another compatibility flag, while we already have `supportedGuestOS`. I'm thinking ahead to the v1alpha2 proposal and trying to do the conversion in a nice way if we do this. 

Thoughts about instead introducing something like strring[] `Incompatibilities` where we can add arbitrary string reasons (same way Conditions are done). Three specific reasons I can already think of are `IncompatibleOsType`, `IncompatibleHwVersion`, `NonVmServiceCompliantImage` (last one is the one that would look for the magic compat key). 

We can optionally also have a `Supported` field which would be equivalent to `len(Incompatibilities)==0`

cc @sreyasn

api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -73,9 +73,24 @@ type VirtualMachineImageStatus struct {
 	// Deprecated
 	PowerState    string `json:"powerState,omitempty"`
 
-	// SupportedGuestOS indicates whether the VirtualMachineImage's osType is supported
-	// by the hosts in the cluster and VM Service
-	SupportedGuestOS *bool  `json:"supportedGuestOS,omitempty"`
+	// ImageSupportedState indicates whether the VirtualMachineImage is supported by VM Service,
+	// A VirtualMachineImage is supported by VM Service if:
+	// - If OS Type is of Linux Family
```
sreyasn, line 78: also indicates support on the hosts of the cluster.

api/v1alpha1/condition_consts.go
```diff
@@ -37,3 +37,13 @@ const (
 	// DeletedReason (Severity=Info) documents a condition not in Status=True because the underlying object was deleted.
 	DeletedReason = "Deleted"
 )
+
+// Conditions related to the VirtualMachineImages
+const(
+	// VirtualMachineImageOsTypeNotSupportedCondition denotes that the OS type in the virtual machine image object is
+	// supported by VMService or not. e.g. Image with windows OS type is not supported by VM Service.
+	VirtualMachineImageOsTypeNotSupportedCondition ConditionType = "OS Type in VirtualMachineImage is not supported"
```
bryanv, line 45: These should be one camel-cased word

api/v1alpha1/virtualmachineimage_types.go
bryanv: This is probably better expressed in terms of conditions are true or not

api/v1alpha1/condition_consts.go
```diff
@@ -37,3 +37,13 @@ const (
 	// DeletedReason (Severity=Info) documents a condition not in Status=True because the underlying object was deleted.
 	DeletedReason = "Deleted"
 )
+
+// Conditions related to the VirtualMachineImages
+const(
+	// VirtualMachineImageOsTypeNotSupportedCondition denotes that the OS type in the virtual machine image object is
```
bryanv, line 43: These should be positive values so don't use Not.

https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20200506-conditions.md
> "Condition types MUST have a consistent polarity (i.e. "True = good");

api/v1alpha1/virtualmachineimage_types.go
srajashekar-vmw: Yes agreed, the conditions have a detailed description of each of these cases, This is the summary field that we discussed in the call the other day, and this documentation useful to understand when this boolean would be set

api/v1alpha1/virtualmachineimage_types.go
srajashekar-vmw: This field is removed and the documentation is also updated

api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -73,9 +73,25 @@ type VirtualMachineImageStatus struct {
...
+	// - If OS Type is supported by hosts in the cluster
+	// - If the image is a photon image or if the image is prepared by VI admin by adding vmware-system.vmsvc.compatibility key
```
janitha, line 80: Let's not call out the explicit key here (if we have to, we will document it elsewhere), but rather simply state "The image has been specifically prepared for VM Service consumption"

api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -73,9 +73,25 @@ type VirtualMachineImageStatus struct {
...
+	ImageSupportedState *bool  `json:"imageSupportedState,omitempty"`
```
janitha, line 81: I don't think the State suffix is needed for this

api/v1alpha1/condition_consts.go
```diff
@@ -37,3 +37,19 @@ const (
...
 	DeletedReason = "Deleted"
 )
+
+// Conditions related to the VirtualMachineImages
+const(
+	// VirtualMachineImageOsTypeSupportedCondition denotes that the OS type in the virtual machine image object is
+	// supported by VMService or not. e.g. Image with windows OS type is not supported by VM Service.
+	VirtualMachineImageOsTypeSupportedCondition ConditionType = "VirtualMachineImageOsTypeSupported"
+	// VirtualMachineImageSupportedCondition denotes that the image compatibility with VMService. VMService expects
+	// VirtualMachineImage to be prepared by VI admin by adding "vmware-system.vmsvc.compatibility" key or a photon image.
```
janitha, line 47: This condition name conflicts with the purpose of `ImageSupportedState` below (which would be a combination or multiple conditions). 

How about this condition be renamed `VirtualMachineImageVmwarePreparedImageCondition` or `VirtualMachineImage

Also remove the note about the actual key and VI Admins, we should just state it's a Image that's specially prepared for VM Service. The decision is still a bit in flux, but initially vmware and related vendors will create the images according to our instructions and publish with the key.

api/v1alpha1/condition_consts.go
```diff
@@ -37,3 +37,21 @@ const (
 	// DeletedReason (Severity=Info) documents a condition not in Status=True because the underlying object was deleted.
 	DeletedReason = "Deleted"
 )
+
+// Conditions related to the VirtualMachineImages
+const(
+	// VirtualMachineImageOsTypeSupportedCondition denotes that the OS type in the virtual machine image object is
+	// supported by VMService or not. e.g. Image with windows OS type is not supported by VM Service.
```
bryanv, line 44: Can we consistently use either "VMService" or "VM Service"?

I think the "or not" can be omitted

"windows" might need to be capitalized, since we do for "Linux" below

api/v1alpha1/condition_consts.go
```diff
@@ -37,3 +37,21 @@ const (
...
+	// VirtualMachineImageSupportedCondition denotes that the image compatibility with VMService. VMService expects
+	// VirtualMachineImage to be prepared by VI admin for VM Service consumption or is a photon image.
+	VirtualMachineImagePreparedForVmserviceCondition ConditionType = "VirtualMachineImagePreparedForVmserviceCondition"
+)
+
+// Condition.Reason for Conditions related to VirtualMachineImages
+const (
+	VirtualMachineImageOsTypeSupportedReason = "VirtualMachineImageOsTypeSupported"
+
+	VirtualMachineImagePreparedForVmserviceReason = "VmServiceCompatibleVirtualMachineImage"
```
bryanv, line 56: "Vmservice" -> "VmService"

api/v1alpha1/virtualmachineimage_types.go
bryanv: We are restating the conditions that need to be true. These descriptions are going to drift, and it already is because the `OsTypeSupportedCondition` condition doesn't say anything about "If OS Type is supported by hosts in the cluster". Express this in the conditions that need to be satisfied because this field is a summary of those.

api/v1alpha1/condition_consts.go
```diff
@@ -37,3 +37,23 @@ const (
...
+	VirtualMachineImageOsTypeSupportedCondition ConditionType = "VirtualMachineImageOsTypeSupported"
+
+	// VirtualMachineImageSupportedCondition denotes image compatibility with VMService. VMService expects
+	// VirtualMachineImage to be prepared by VI admin for VMService consumption or is a photon image.
+	VirtualMachineImagePreparedForVmServiceCondition ConditionType = "VirtualMachineImagePreparedForVmServiceCondition"
+)
+
+// Condition.Reason for Conditions related to VirtualMachineImages
+const (
+	VirtualMachineImageOsTypeSupportedReason = "VirtualMachineImageOsTypeSupported"
```
aruneshpa, line 56: IMHO, we should not have to set `Reason` for a True condition type. 
https://github.com/kubernetes-sigs/cluster-api/blob/master/util/conditions/setter.go#L109


We can rephrase these (maybe 
`VirtualMachineImageOsTypeUnSupportedReason` and `UnsupportedVirtualmachineImage`) such that we only set these when we have to mark the above conditions as False.

srajashekar-vmw: https://github.com/vmware-tanzu/vm-operator-api/pull/55#discussion_r568063297
I will let you and @bryanv come to a consensus on what convention to follow

srajashekar-vmw: fixed

api/v1alpha1/condition_consts.go
```diff
@@ -37,3 +37,23 @@ const (
 	// DeletedReason (Severity=Info) documents a condition not in Status=True because the underlying object was deleted.
 	DeletedReason = "Deleted"
 )
+
+// Conditions related to the VirtualMachineImages
+const(
+	// VirtualMachineImageOsTypeSupportedCondition denotes that the OS type in the virtual machine image object is
+	// supported by VMService. A VirtualMachineImageOsTypeSupportedCondition is marked true if
```
bryanv, line 44: Probably don't need `if` here, and to start each item below with `If` too

api/v1alpha1/virtualmachineimage_types.go
srajashekar-vmw: Fixed

api/v1alpha1/condition_consts.go
```diff
@@ -37,3 +37,23 @@ const (
 	// DeletedReason (Severity=Info) documents a condition not in Status=True because the underlying object was deleted.
 	DeletedReason = "Deleted"
 )
+
+// Conditions related to the VirtualMachineImages
+const(
+	// VirtualMachineImageOsTypeSupportedCondition denotes that the OS type in the virtual machine image object is
```
aruneshpa, line 43: Here and below:
https://github.com/golang/go/wiki/CodeReviewComments#initialisms
`OsType`-> `OSType`

Also, `virtual machine image object` -> `VirtualMachineImage object`

api/v1alpha1/condition_consts.go
```diff
@@ -37,3 +37,23 @@ const (
...
+	// supported by VMService. A VirtualMachineImageOsTypeSupportedCondition is marked true
+	// - If OS Type is of Linux Family
+	// - If OS Type is supported by hosts in the cluster
+	VirtualMachineImageOsTypeSupportedCondition ConditionType = "VirtualMachineImageOsTypeSupported"
+
+	// VirtualMachineImageSupportedCondition denotes image compatibility with VMService. VMService expects
+	// VirtualMachineImage to be prepared by VI admin for VMService consumption or is a photon image.
+	VirtualMachineImagePreparedForVmServiceCondition ConditionType = "VirtualMachineImagePreparedForVmServiceCondition"
```
aruneshpa, line 51: `VmServiceCondition` -> `VMServiceCondition`

api/v1alpha1/condition_consts.go
```diff
@@ -37,3 +37,23 @@ const (
...
+)
+
+// Condition.Reason for Conditions related to VirtualMachineImages
+const (
+	VirtualMachineImageOsTypeNotSupportedReason = "VirtualMachineImageOsTypeNotSupported"
```
aruneshpa, line 56: Here and below, maybe add a comment indicating what this means? This is self explanatory, but it would help with the documentation generation.

api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -73,9 +73,24 @@ type VirtualMachineImageStatus struct {
 	// Deprecated
 	PowerState    string `json:"powerState,omitempty"`
 
-	// SupportedGuestOS indicates whether the VirtualMachineImage's osType is supported
-	// by the hosts in the cluster and VM Service
-	SupportedGuestOS *bool  `json:"supportedGuestOS,omitempty"`
+	// ImageSupported indicates whether the VirtualMachineImage is supported by VMService,
```
aruneshpa, line 76: Comma -> Period at the EOL?

aruneshpa (approved):

api/v1alpha1/condition_consts.go
```diff
@@ -37,3 +37,23 @@ const (
...
+
+// Conditions related to the VirtualMachineImages
+const(
+	// VirtualMachineImageOsTypeSupportedCondition denotes that the OS type in the virtual machine image object is
+	// supported by VMService. A VirtualMachineImageOsTypeSupportedCondition is marked true
+	// - If OS Type is of Linux Family
+	// - If OS Type is supported by hosts in the cluster
+	VirtualMachineImageOsTypeSupportedCondition ConditionType = "VirtualMachineImageOsTypeSupported"
+
+	// VirtualMachineImageSupportedCondition denotes image compatibility with VMService. VMService expects
```
janitha, line 49: Now we have better clarity on path forward, could you rename this to `VirtualMachineImagev1alpha1CompatibleCondition` to be more specific. And the description should state the VirtualMachinieImage will be prepared by VMware specifically for VM Service v1alpha1. (and drop the bit about Photon, for v1alpha1 we will be preparing custom Photon images as well). 

Sorry for renaming churn :(

api/v1alpha1/condition_consts.go
```diff
@@ -37,3 +37,23 @@ const (
...
+	// VirtualMachineImage to be prepared by VI admin for VMService consumption or is a photon image.
+	VirtualMachineImagePreparedForVmServiceCondition ConditionType = "VirtualMachineImagePreparedForVmServiceCondition"
+)
+
+// Condition.Reason for Conditions related to VirtualMachineImages
+const (
+	VirtualMachineImageOsTypeNotSupportedReason = "VirtualMachineImageOsTypeNotSupported"
+
+	VirtualMachineImageNotPreparedForVmServiceReason = "VirtualMachineImageNotPreparedForVmService"
```
janitha, line 58: Update to reflect `VirtualMachineImagev1alpha1CompatibleCondition`

janitha (approved):

api/v1alpha1/condition_consts.go
srajashekar-vmw: fixed

api/v1alpha1/condition_consts.go
srajashekar-vmw: updated to `VirtualMachineImageV1Alpha1CompatibleCondition` to reflect Janitha's comment

janitha (approved):
Awesome sauce

api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -92,9 +92,24 @@ type VirtualMachineImageStatus struct {
 	// Deprecated
 	PowerState    string `json:"powerState,omitempty"`
 
-	// SupportedGuestOS indicates whether the VirtualMachineImage's osType is supported
-	// by the hosts in the cluster and VM Service
-	SupportedGuestOS *bool  `json:"supportedGuestOS,omitempty"`
+	// ImageSupported indicates whether the VirtualMachineImage is supported by VMService.
+	// A VirtualMachineImage is supported by VMService if the following conditions are true:
+	// - VirtualMachineImageOSTypeSupportedCondition
+	// - VirtualMachineImageV1Alpha1CompatibleCondition
```
bryanv, line 98: This should have `// +optional` too

api/v1alpha1/condition_consts.go
```diff
@@ -49,3 +49,27 @@ const (
...
+// Conditions related to the VirtualMachineImages
+const(
+	// VirtualMachineImageOSTypeSupportedCondition denotes that the OS type in the VirtualMachineImage object is
+	// supported by VMService. A VirtualMachineImageOsTypeSupportedCondition is marked true:
+	// - If OS Type is of Linux Family
+	// - If OS Type is supported by hosts in the cluster
+	VirtualMachineImageOSTypeSupportedCondition ConditionType = "VirtualMachineImageOSTypeSupported"
+
+	// VirtualMachineImageV1Alpha1CompatibleCondition denotes image compatibility with VMService. VMService expects
+	// VirtualMachineImage to be prepared by VMware specifically for VM Service v1alpha1.
```
bryanv, line 62: "VM Service" -> "VMService" for consistency?

srajashekar-vmw: I had changed all of these to VMService, rebase probably messed this up. Thanks for pointing it out. Will fix it

PR #59, opened 2021-02-04
Enforce enum restrictions on VM power state and metadata transport 

aruneshpa:
The power state of the VirtualMachine from the spec should match exactly as the
power states defined in the vim 25 API descriptions. The only allowed values are
"powredOn" and "poweredOff".

Similarly, the supported values for VM metadata transport are OvfEnv and
ExtraConfig, but we do not enforce them.

This change enforces these values by leveraging the kubebuilder validation
markers.

### DIFF
--- a/api/v1alpha1/virtualmachine_types.go
+++ b/api/v1alpha1/virtualmachine_types.go
@@ -9,13 +9,16 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-type VirtualMachinePowerState string
-
 // See govmomi.vim25.types.VirtualMachinePowerState
 const (
 	VirtualMachinePoweredOff VirtualMachinePowerState = "poweredOff"
 	VirtualMachinePoweredOn  VirtualMachinePowerState = "poweredOn"
 )
+
+// VirtualMachinePowerState represents the power state of a VirtualMachine.
+// The value values are "poweredOn", and "poweredOff".
+// +kubebuilder:validation:Enum=poweredOff;poweredOn
+type VirtualMachinePowerState string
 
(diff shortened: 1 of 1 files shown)


janitha:
Pretty cool to know about `kubebuilder:validation:Enum`

Can we go ahead and do this for all the enumy variables? eg. `VMStatusPhase` `VirtualMachineMetadataTransport` and there maybe more in other CRs, I only briefly looked.

janitha (approved):

api/v1alpha1/virtualmachine_types.go
```diff
@@ -9,14 +9,17 @@ import (
...
-
 // See govmomi.vim25.types.VirtualMachinePowerState
 const (
 	VirtualMachinePoweredOff VirtualMachinePowerState = "poweredOff"
 	VirtualMachinePoweredOn  VirtualMachinePowerState = "poweredOn"
 )
 
+// VirtualMachinePowerState represents the power state of a VirtualMachine.
+// The value values are "poweredOn", and "poweredOff".
+// +kubebuilder:validation:Enum=poweredOff;poweredOn
```
bryanv, line 20: I think today we have situations where the status field will be the default value (""), and it isn't clear to me that this would allow that.

aruneshpa: I tried out the service with this change. As of now, we do not update the power state with "" anywhere. We either don't fill it out in the `VirtualMachineStatus`, or fill valid values (`poweredOff`, `poweredOn`). As long as we continue to do that, we won't have any issues.

aruneshpa: I am marking this as resolved.

PR #60, opened 2021-02-06
Expose ovfEnv properties in VirtualMachineImage

yi0909:
Introduce an OVF properties field to the VirtualMachineImange spec, 
so that a developer can do a describe on the virtualmachineimage and 
see ovf properties.

Properties with false ovf:userConfigurable value will be omit. 
A developer can only see the userConfigurable key, type and value.


api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -41,6 +41,21 @@ type VirtualMachineImageOSInfo struct {
...
+type VirtualMachineImageOVFEnv struct {
+	// Key describes the key of the ovf property.
+	Key string `json:"key"`
+
+	// Type describes the type of the ovf property.
+	Type string `json:"type"`
+
+	// Value describes the default value of the ovf key.
+	// +optional
+	Value *string `json:"value,omitempty"`
```
janitha, line 56: Rename this to `Default` (value is what the user will provide)

janitha (approved):

api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -41,6 +41,21 @@ type VirtualMachineImageOSInfo struct {
 	Type string `json:"type,omitempty"`
 }
 
+
+// VirtualMachineImageOVFEnv describes information related to a user configurable property element
+// that is supported by VirtualMachineImage and can be customized during VirtualMachine deployment.
```
bryanv, line 46: Don't use "deployment" because that means something very different in kubernetes

yi0909: Got it. I will fix this in another MR.

PR #62, opened 2021-03-04
Update comment on extraConfig transport stating only guestinfo* prefixed keys to be allowed

janitha:
Update comment on extraConfig transport stating only guestinfo* prefixed keys to be allowed

### DIFF
--- a/api/v1alpha1/virtualmachine_types.go
+++ b/api/v1alpha1/virtualmachine_types.go
@@ -98,7 +98,7 @@ type VirtualMachineMetadataTransport string
 
 const (
 	// VirtualMachineMetadataExtraConfigTransport will set the VirtualMachineMetadata ConfigMap data as
-	// extraConfig key value fields on the VM.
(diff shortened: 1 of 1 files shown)


api/v1alpha1/virtualmachine_types.go
```diff
@@ -98,7 +98,7 @@ type VirtualMachineMetadataTransport string
 
 const (
 	// VirtualMachineMetadataExtraConfigTransport will set the VirtualMachineMetadata ConfigMap data as
-	// extraConfig key value fields on the VM.
+	// extraConfig key value fields on the VM. Only keys prefixed with "guestinfo." will be allowed.
```
bryanv, line 101: I'm not sure if "allowed" is the right wordage here since there is no feedback. Maybe "set" or "copied" instead?

janitha: Ah yeah, I'll change it to "set"

bryanv (approved):

PR #63, opened 2021-03-31
Add provision to specify GPU devices in VirtualMachineClass

jainvarun21:
This change adds provisions to specify vGPU and Dynamic DirectPath I/O
(used for configuring devices in passthrough mode) devices in
VirtualMachineClass.

Testing Done:
- make all


api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -8,10 +8,29 @@ import (
...
+// VGPUDevice contains the configuration corresponding to a vGPU device.
+type VGPUDevice struct {
+	ProfileName string `json:"profileName,omitempty"`
+}
+
+// DynamicDirectPathIODevice contains the configuration corresponding to a Dynamic DirectPath I/O device.
+type DynamicDirectPathIODevice struct {
+	VendorId    int    `json:"vendorId,omitempty"`
+	DeviceId    int    `json:"deviceId,omitempty"`
+	CustomLabel string `json:"customLabel,omitempty"`
```
janitha, line 20: Are all three fields required here? I thought CustomLabel was optional (or vice-versa)

api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -8,10 +8,29 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// VGPUDevice contains the configuration corresponding to a vGPU device.
+type VGPUDevice struct {
+	ProfileName string `json:"profileName,omitempty"`
```
bryanv, line 13: Being the only field, I don't think `omitempty` make sense

api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -8,10 +8,29 @@ import (
...
+}
+
+// DynamicDirectPathIODevice contains the configuration corresponding to a Dynamic DirectPath I/O device.
+type DynamicDirectPathIODevice struct {
+	VendorId    int    `json:"vendorId,omitempty"`
```
bryanv, line 18: `ID` instead of `Id` matches k8s convention

Does `omitempty` really apply to all of these?

api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -8,10 +8,29 @@ import (
...
+	DeviceId    int    `json:"deviceId,omitempty"`
+	CustomLabel string `json:"customLabel,omitempty"`
+}
+
+// VirtualDevices contains information about the virtual devices associated with a VirtualMachineClass.
+type VirtualDevices struct {
+	VgpuDevices                []VGPUDevice                `json:"vgpuDevices,omitempty" patchStrategy:"merge" patchMergeKey:"profileName"`
```
bryanv, line 25: `VGPUDevices`?

I think these should have `// +optional`

api/v1alpha1/virtualmachineclass_types.go
jainvarun21: You're right. CustomLabel is optional.

api/v1alpha1/virtualmachineclass_types.go
jainvarun21: You're right. I removed them now.

janitha (approved):

api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -8,10 +8,33 @@ import (
...
+}
+
 // VirtualMachineClassHardware describes a virtual hardware resource specification.
 type VirtualMachineClassHardware struct {
-	Cpus   int64             `json:"cpus,omitempty"`
-	Memory resource.Quantity `json:"memory,omitempty"`
+	Cpus    int64             `json:"cpus,omitempty"`
+	Memory  resource.Quantity `json:"memory,omitempty"`
+	// +optional
+	Devices VirtualDevices    `json:"devices,omitempty"`
```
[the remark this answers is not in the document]
bryanv: I kind of prefer that as well. It also matches how we ended up doing the `VirtualMachineVolume`. We'll need to ensure we have a validation webhook to force proper usage.

PR #64, opened 2021-04-01
Remove client-gen and associated bits

janitha:
Client-gen is not used anymore, removing it plus everything else associated with it.

Testing done: 
`make all`
`make -C hack/samples all`


hack/tools/Makefile
```diff
@@ -18,9 +18,6 @@ HOST_ARCH=$(shell go env GOARCH)
 # Binaries.
 CONTROLLER_GEN  := $(BIN_DIR)/controller-gen
 GOLANGCI_LINT   := $(BIN_DIR)/golangci-lint
-CLIENT_GEN      := $(BIN_DIR)/client-gen
```
bryanv, line 21: I think `_ "k8s.io/code-generator"` can be removed from tools.go now, and then run `go mod tidy` to remove that dependency

janitha: Ah yeah, good call

bryanv (approved):

sreyasn (approved):

PR #66, opened 2021-04-22
Add GuestHeartbeat probe action

bryanv:
This uses the "guestHeartBeatStatus" field in the VM VMODL to define a
new probe type that does not depend on network connectivity between the
supervisor CP VMs and the TKG or VMService VMs.


janitha (approved):

api/v1alpha1/virtualmachine_types.go
```diff
@@ -185,6 +189,26 @@ type TCPSocketAction struct {
 	Host string `json:"host,omitempty"`
 }
 
+// The guest heartbeat status.
+type GuestHeartbeatStatus string
+
+// See govmomi.vim25.types.ManagedEntityStatus
+const (
+	// Intermittent heartbeat. May be due to guest load.
+	YellowStatus GuestHeartbeatStatus = "yellow"
```
aruneshpa, line 198: Should we change the name to reflect that its related to Heartbeat? Seeing `YellowStatus` might not make it clear that we are talking about Heartbeat status (e.g., it is used in quite a few places in vim25/types/types.go). 

What about renaming it to "YellowHeartbeat"?

api/v1alpha1/virtualmachine_types.go
```diff
@@ -185,6 +189,26 @@ type TCPSocketAction struct {
...
+	// Guest operating system is responding normally.
+	GreenStatus  GuestHeartbeatStatus = "green"
+)
+
+// GuestHeartbeatAction describes an action based on the guest heartbeat.
+type GuestHeartbeatAction struct {
```
aruneshpa, line 204: The name `GuestHeartbeatAction` feels like we will take certain actions based on the heartbeat of the Guest. But we are not doing that here. Maybe we can call it `GuestHeartbeatProbe` since this behaves more like a probe?

api/v1alpha1/virtualmachine_types.go
```diff
@@ -185,6 +189,26 @@ type TCPSocketAction struct {
...
+	// MinimumStatus is the minimum heartbeat status that is considered successful.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=green
+	// +kubebuilder:validation:Enum=yellow;green
+	MinimumStatus GuestHeartbeatStatus `json:"minimumStatus,omitempty"`
```
aruneshpa, line 209: What do you think about calling it `ThresholdStatus` and adding a comment that says whenever the heartbeat status doesn't pass the threshold, the probe is considered to fail?

janitha (approved):
Could you expand the comment on `Probe` to clarify that either TCPSocket or GuestHeartbeat, and not both, can be set (this wasn't obvious to me until I was looking at your other MR)

api/v1alpha1/virtualmachine_types.go
bryanv: `Action` is based on the existing `TcpSocketAction` name. I'm not really a fan of that name but think we should keep consistency.

api/v1alpha1/virtualmachine_types.go
bryanv: That's fine

api/v1alpha1/virtualmachine_types.go
bryanv: Sure

api/v1alpha1/virtualmachine_types.go
aruneshpa: Makes sense to me.

aruneshpa (approved):

PR #67, opened 2021-04-22
Support Multiple NICs in VirtualMachine Status

sreyasn:
This change adds support for multiple network interfaces in
the VirtualMachine Status.

testing done:
make-all


api/v1alpha1/virtualmachine_types.go
```diff
@@ -281,6 +281,20 @@ type VirtualMachineVolumeStatus struct {
 	Error string `json:"error"`
 }
 
+// NetworkInterfaceStatus defines the observed state of the network interfaces desired to be attached
```
janitha, line 284: I would reword this comment to clarify these are the interfaces as seen by vmtools and the guest os. 

(the desired interfaces would be configured in the spec, but it's possible the os would have ignored them, specially in future cases where additional images are supported)

api/v1alpha1/virtualmachine_types.go
```diff
@@ -299,8 +313,7 @@ type VirtualMachineStatus struct {
 	// +optional
 	Conditions []Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
 
-	// VmIp describes the IP address of the VirtualMachine.  Currently, a VirtualMachine only supports a single
-	// network interface and interface address.
+	// VmIp describes the IP address of the VirtualMachine.
```
janitha, line 316: Possibly clarify that multiple IPs can be available for the VM, and to see the NetworkInterfaces for additional ones.

janitha (approved):

api/v1alpha1/virtualmachine_types.go
sreyasn: sounds good. To verify, Is it okay for us to mention VMware tools in the comments here?.

janitha: Yeah, I don't see any problem with that

api/v1alpha1/virtualmachine_types.go
bryanv: > Possibly clarify that multiple IPs can be available for the VM, and to see the NetworkInterfaces for additional ones.

The VMODL docs that correspond is 
> Primary IP address assigned to the guest operating system, if known.

I'd make some mention of that to help users better understand what this IP is vs the others.

api/v1alpha1/virtualmachine_types.go
bryanv: I think later we'll have to incorporate and match these up with their corresponding spec entry when possible

bryanv (approved):

PR #73, opened 2021-05-17
Add Virtual Hardware Version to the VirtualMachineImage Spec and some cleanup

sreyasn:
- This change adds the ovf Hardware Version to the VirtualMachineImage
object.
- It also deprecates some condition constants which won't be used

### DIFF
--- a/api/v1alpha1/condition_consts.go
+++ b/api/v1alpha1/condition_consts.go
@@ -52,6 +52,7 @@ const (
 
 // Conditions related to the VirtualMachineImages
 const(
+	// Deprecated
 	// VirtualMachineImageOSTypeSupportedCondition denotes that the OS type in the VirtualMachineImage object is
 	// supported by VMService. A VirtualMachineImageOsTypeSupportedCondition is marked true:
 	// - If OS Type is of Linux Family
@@ -65,6 +66,7 @@ const(
 
 // Condition.Reason for Conditions related to VirtualMachineImages
 const (
+	// Deprecated
 	// VirtualMachineImageOSTypeNotSupportedReason (Severity=Error) documents that OS Type is VirtualMachineImage is
 	// not supported.
 	VirtualMachineImageOSTypeNotSupportedReason = "VirtualMachineImageOSTypeNotSupported"
--- a/api/v1alpha1/virtualmachineimage_types.go
+++ b/api/v1alpha1/virtualmachineimage_types.go
@@ -78,6 +78,10 @@ type VirtualMachineImageSpec struct {
 	// OVFEnv describes the user configurable customization parameters of the VirtualMachineImage.
 	// +optional
 	OVFEnv map[string]OvfProperty `json:"ovfEnv,omitempty"`
+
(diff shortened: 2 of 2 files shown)


api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -78,6 +78,10 @@ type VirtualMachineImageSpec struct {
 	// OVFEnv describes the user configurable customization parameters of the VirtualMachineImage.
 	// +optional
 	OVFEnv map[string]OvfProperty `json:"ovfEnv,omitempty"`
+
+	// HardwareVersion describes the virtual hardware version of the image
+	// +optional
+	HardwareVersion int32 `json:"hwVersion,omitempty"`
```
janitha, line 84: To make this optional, shouldn't this be a `*int32` (otherwise unset value would equal 0)

api/v1alpha1/condition_consts.go
```diff
@@ -52,23 +52,13 @@ const (
 
 // Conditions related to the VirtualMachineImages
 const(
-	// VirtualMachineImageOSTypeSupportedCondition denotes that the OS type in the VirtualMachineImage object is
```
janitha, line 55: Add a comment stating deprecated, but don't remove them (all changes are meant to be additive), and these have been here for a while now. 

Same for the other stuff below.

api/v1alpha1/virtualmachineimage_types.go
sreyasn: I had that question too. I found out that when the value is unset 0 (ie) empty for an int32, it gets omitted from describe/get for the virtualmachineimage object. And since we'll be the only consumers right now, All checks in vmoperator look for unset val 0 before making any decisions based on it.

janitha: Ah okay, we can keep it as is, thanks for checking

api/v1alpha1/condition_consts.go
```diff
@@ -52,10 +52,7 @@ const (
...
-	// supported by VMService. A VirtualMachineImageOsTypeSupportedCondition is marked true:
-	// - If OS Type is of Linux Family
-	// - If OS Type is supported by hosts in the cluster
```
aruneshpa, line 58: Is there a reason we are removing these comments? Just adding the `//Deprecated` should be enough?

aruneshpa (approved):

api/v1alpha1/condition_consts.go
sreyasn: yeah. you're probably right.

sreyasn: My rationale was why try to describe something that is deprecated. I also remember doing something similar to a few status fields in the VirtualMachineImage.

janitha (approved):

api/v1alpha1/condition_consts.go
aruneshpa: I think we can leave the comments. Would hurt the readability to have a field without a description.

PR #76, opened 2021-06-16
Introduce VSphereVMCustomization condition and reasons

janitha:
This introduces VSphereVMCustomization condition and reasons that expose the current status (weather it's running/pending/success/fail).

The status maps 1:1 with the recently released (7.0.2.0) vSphere GuestInfo API CustomizationInfo

### DIFF
--- a/api/v1alpha1/condition_consts.go
+++ b/api/v1alpha1/condition_consts.go
@@ -35,6 +35,24 @@ const (
 	// VirtualMachineImageNotFoundReason (Severity=Error) documents that the VirtualMachineImage specified in the VirtualMachineSpec
 	// is not available.
 	VirtualMachineImageNotFoundReason = "VirtualMachineImageNotFound"
+
+	// VSphereVMCustomizationCondition exposes the status of vSphere VM Customization from within the guest OS, when available.
+	VSphereVMCustomizationCondition ConditionType = "VSphereVMCustomization"
+
+	// VSphereVMCustomizationIdleReason (Severity=Info) documents that vSphere VM Customizations were not applied for the VirtualMachine.
+	VSphereVMCustomizationIdleReason = "VSphereVMCustomizationIdle"
+
+	// VSphereVMCustomizationPendingReason (Severity=Info) documents that vSphere VM Customization is still pending within the guest OS.
+	VSphereVMCustomizationPendingReason = "VSphereVMCustomizationPending"
+
+	// VSphereVMCustomizationRunningReason (Severity=Info) documents that the vSphere VM Customization is now running on the guest OS.
+	VSphereVMCustomizationRunningReason = "VSphereVMCustomizationRunning"
+
+	// VSphereVMCustomizationSucceededReason (Severity=Info) documents that the vSphere VM Customization succeeded within the guest OS.
+	VSphereVMCustomizationSucceededReason = "VSphereVMCustomizationSucceeded"
+
+	// VSphereVMCustomizationFailedReason (Severity=Error) documents that the vSphere VM Customization failed within the guest OS.
+	VSphereVMCustomizationFailedReason = "VSphereVMCustomizationFailed"
 )
 
 // Common Condition.Reason used by VM Operator API objects.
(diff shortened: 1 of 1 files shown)


api/v1alpha1/condition_consts.go
```diff
@@ -35,6 +35,24 @@ const (
 	// VirtualMachineImageNotFoundReason (Severity=Error) documents that the VirtualMachineImage specified in the VirtualMachineSpec
 	// is not available.
 	VirtualMachineImageNotFoundReason = "VirtualMachineImageNotFound"
+
+	// VSphereVMCustomizationCondition exposes the status of vSphere VM Customization from within the guest OS, when available.
```
akutz, line 39: Hi @janitha,

Is there a reason the reasons are not an aliased `string` type as well? Kubernetes API conventions [prefers](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types) specific types to `string` type when possible:

> Use aliases for string instead (e.g., `NodeConditionType`).

Or does the condition itself require a `string`? Even if it does, having an aliased type makes testing a little easier.

FWIW, I did [double check](https://play.golang.org/p/Fkt01ttt6Bo), and while I thought Golang would continue to set the type of constant values to the prior, explicit type, this does not seem to be the case:

```bash
VSphereVMCustomization             main.ConditionType
VSphereVMCustomizationIdle         string
VSphereVMCustomizationPending      string

Program exited.
```

akutz (approved):
Looks good to me, thank you! I left one comment, but please do not consider it blocking.

api/v1alpha1/condition_consts.go
janitha: Not really sure about the history/reasonings behind this pattern, perhaps @bryanv can shed some light. 

Looking at other k8s projects, even corev1, the pattern seems to be the same as this, where condition is typed alias, but the reason/message are optional strings https://pkg.go.dev/k8s.io/api/core/v1#NamespaceCondition

I think if we were to update/clean this up, we should do it as a separate MR, and also do the changes on the implementation side.

akutz: Sure thing. As I said, the type alias is really only a thing when it's a proper field, [for example](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/34bbb00dff743efab8b8e149dcf07d124f716fd6/api/v1alpha3/types.go#L36-L68):

```golang
// CloneMode is the type of clone operation used to clone a VM from a template.
type CloneMode string

const (
	// FullClone indicates a VM will have no relationship to the source of the
	// clone operation once the operation is complete. This is the safest clone
	// mode, but it is not the fastest.
	FullClone CloneMode = "fullClone"

	// LinkedClone means resulting VMs will be dependent upon the snapshot of
	// the source VM/template from which the VM was cloned. This is the fastest
	// clone mode, but it also prevents expanding a VMs disk beyond the size of
	// the source VM/template.
	LinkedClone CloneMode = "linkedClone"
)

// VirtualMachineCloneSpec is information used to clone a virtual machine.
type VirtualMachineCloneSpec struct {
	// Template is the name or inventory path of the template used to clone
	// the virtual machine.
	// +kubebuilder:validation:MinLength=1
	Template string `json:"template"`


	// CloneMode specifies the type of clone operation.
	// The LinkedClone mode is only support for templates that have at least
	// one snapshot. If the template has no snapshots, then CloneMode defaults
	// to FullClone.
	// When LinkedClone mode is enabled the DiskGiB field is ignored as it is
	// not possible to expand disks of linked clones.
	// Defaults to LinkedClone, but fails gracefully to FullClone if the source
	// of the clone operation has no snapshots.
	// +optional
	CloneMode CloneMode `json:"cloneMode,omitempty"`

	...
```

There it made sense. But as a value for a `string` in a generic `Condition` used by multiple consumers it makes less sense.

bryanv: The short reason would be that all the Conditions code here is from cluster-api - with the minimum search & replace applied to adopt it for VM Operator - where the `Reason` is a `string`. As @janitha mentioned, my guess would be that cluster-api used `string` because that is what seems to be used everywhere in corev1, and consistently won over convention.

jainvarun21:
Looks good to me.

PR #77, opened 2021-06-29
Add ImageId and ImageInfo to VirtualMachineImage

yi0909:
- This change adds ImageId to VirtualmachineImage, which tells the item UUID in the vCenter.
- Add ImageInfo to VirtualMachineImage, which includes this item display name and its content
  library UUID. This can help to support duplicate VM image names in the cluster.
- Print the content source UUID when doing kubectl get vmimage.


api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -119,6 +134,7 @@ func (vmImage *VirtualMachineImage) SetConditions(conditions Conditions) {
 // +kubebuilder:resource:scope=Cluster,shortName=vmimage
 // +kubebuilder:storageversion
 // +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="ImageSourceId",type="string",JSONPath=".status.imageInfo.imageSourceId"
```
janitha, line 137: The imageSourceId is a uuid, which isn't really human friendly (and kubebuilder:printcolumn:name are meant for human consumption). Perhaps the ImageName would be better here, thoughts?

api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -41,6 +41,14 @@ type VirtualMachineImageOSInfo struct {
 	Type string `json:"type,omitempty"`
 }
 
+type VirtualMachineImageInfo struct {
+	// ImageName describes the display name of this VirtualMachineImage.
+	ImageName string `json:"imageName,omitempty"`
+
+	// ImageSourceID describes the uuid of the content library which owns this VirtualMachineImage.
+	ImageSourceID string `json:"imageSourceId,omitempty"`
```
janitha, line 49: Should we include a ImageSourceName as well (display name of the content library?)

api/v1alpha1/virtualmachineimage_types.go
yi0909: IMO, for now developers can only know the UUID of which content source they have access to. Maybe print out the ImageSourceId can help developers figure out what vm images belong to the attached content source? 
We only append several random bits to the vm image name, so that it should not lose much readability.  Will it be a little bit verbose if we add another column to print the ImageName?

janitha: Ah okay, so this column is meant for the vmimage's content source. Even in that case, knowing a uuid still isn't all that useful. How about introducing status.imageInfo.imageSourceName (friendly content library name?), and then displaying that instead of the uuid. (while still keeping status.imageinfo.imageSourceId, but not displaying it as a column)

yi0909: Yes a uuid is not human friendly. :(  But we'll have some issues if we don't print the uuid but print the status.imageInfo.imageSourceName instead.
1.  A developer can't match the content source k8s resource name (uuid) and its display name. (This can be done by adding  content source display name field to ContentSourceBinding and ContentSource CRD. )
2. We can't ensure that a VI admin won't create two content libraries with the same name. This scenario is allowed in the vCenter, while UUID is guaranteed to be unique.

api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -41,6 +41,26 @@ type VirtualMachineImageOSInfo struct {
...
 }
 
+// VirtualMachineImageInfo describes basic information related to the image, such as id and name. This information
+// can be used to identify an image in the content source.
+type VirtualMachineImageInfo struct {
+	//ImageID describes the ID of this VirtualMachineImage.
+	ImageID string `json:"imageId"`
+
+	// ImageName describes the display name of this VirtualMachineImage.
+	ImageName string `json:"imageName,omitempty"`
```
bryanv, line 51: What situation would this be empty? Generally, such fields also need to be `// +optional` but I'm not sure why this would be `omitempty`

api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -41,6 +41,26 @@ type VirtualMachineImageOSInfo struct {
...
+}
+
+// VirtualMachineImageSourceInfo describes the uuid and display name related to the source of this image.
+type VirtualMachineImageSourceInfo struct {
```
bryanv, line 55: I think there ends up being a little stuttering with the `VirtualMachineImageSourceInfo` name, and then using `ImageSource` as the prefix for the members. Also, already in the top-level Spec we have a `ImageSourceType` so we've spread out the source info.

Also, seems this should be more like a `ContentSourceRef` because just the `ImageSourceID` isn't necessarily sufficient to identify the source. This`ImageSourceID` and the `Spec.ImageSourceType` tuple may be but how they are grouped is weird. 

A few options might be 
 - Just inline the fields here in the Spec along side the existing `ImageSourceID`
 - Deprecate `Spec.ImageSourceID` and move it here
 - Introduce a full Ref struct here to point to the content library. I think today we end up using the OwnerRefs to infer that.

@aruneshpa should chime in

api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -41,6 +41,26 @@ type VirtualMachineImageOSInfo struct {
 	Type string `json:"type,omitempty"`
 }
 
+// VirtualMachineImageInfo describes basic information related to the image, such as id and name. This information
```
aruneshpa, line 44: In future when VM Service supports other image formats, what would be the value of `id` here? If this is a unique identifier provided by the content provider, we should indicate that here.

api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -41,6 +41,26 @@ type VirtualMachineImageOSInfo struct {
...
+// can be used to identify an image in the content source.
+type VirtualMachineImageInfo struct {
+	//ImageID describes the ID of this VirtualMachineImage.
```
aruneshpa, line 47: Fix the spacing?
`//ImageID` -> `// ImageID`

api/v1alpha1/virtualmachineimage_types.go
aruneshpa: I agree with the stuttering here. `ImageSouurceInfo.ImageSource*` is not ideal. My preference would be to go with something like: `VirtualMachineImageSpec.Soruce.ID`. 

>  seems this should be more like a ContentSourceRef because just the ImageSourceID isn't necessarily sufficient to identify the source.

Yes. It will be complicated to identify a source just by looking at the ID (since we don't include a `type` here). I think you should consider including a `ContentProviderRef` here, since that has all the necessary info like the ID and the type. We can  modify the ContentProvider type to contain the display name as well in the `Status`.  And, the `VirtualMachineImage` to `ContentProvider` relation is expected and well defined.

api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -41,6 +41,16 @@ type VirtualMachineImageOSInfo struct {
 	Type string `json:"type,omitempty"`
 }
 
+// VirtualMachineImageInfo describes basic information related to the image, such as id and name. This information
+// can be used to identify an image in the content source.
+type VirtualMachineImageInfo struct {
```
bryanv, line 46: Other should chime in but my preference would be to move these two fields into the Spec. Arguably the whole Spec is already "info", and I think somebody looking at the currently proposed Spec would wonder why other "info" - OSInfo, ProductInfo, and the other Spec fields - aren't under this "info".

api/v1alpha1/contentlibraryprovider_types.go
```diff
@@ -17,6 +17,9 @@ type ContentLibraryProviderSpec struct {
 // ContentLibraryProviderStatus defines the observed state of ContentLibraryProvider
 // Can include fields indicating when was the last time VM images were updated from a library
 type ContentLibraryProviderStatus struct {
+	// DisplayName describes the display name of the ContentLibraryProvider, which is the name of the vSphere content library.
+	// +optional
+	DisplayName string `json:"displayName,omitempty"`
```
janitha, line 22: Let's do the DisplayName of the ContentLIbrary as a followup story, so this can be removed

api/v1alpha1/virtualmachineimage_types.go
janitha: I think `ImageID` should definitely be in spec, since it's immutable and will be used to uniquely correlate the vmimage with cl item. (this is what the design doc also specified)

However, I think `ImageName` is fine to be in status (since it can change over time). But you're right, the OSInfo/ProductInfo should have probably been done in status as well (since they can also change over time, in case vi admin uploads a new version of the CL Item).

api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -41,6 +41,16 @@ type VirtualMachineImageOSInfo struct {
...
+	// ImageID describes a unique identifier provided by the provider of this VirtualMachineImage.
+	ImageID string `json:"imageId"`
```
bryanv, line 48: To match k8s convention the JSON fields should use "ID", like "imageID" for this field.

api/v1alpha1/virtualmachineimage_types.go
bryanv: I agree that we erred and much of current Spec should have been placed in the Status. That said, I think my preference for now would be to to group similar things together, even if that isn't preferred location. I totally see the other side of let's add new fields to the preferred location going forward, so I don't either have a strong option. Although IDK for a later v1a2 the conversion hooks become a lot more complicated - or just not possible - if moving a field between the Spec and Status subresource.

api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -95,6 +101,9 @@ type VirtualMachineImageStatus struct {
 	// Deprecated
 	PowerState string `json:"powerState,omitempty"`
 
+	// ImageName describes the display name of this VirtualMachineImage.
+	ImageName string `json:"imageName"`
```
bryanv, line 105: I think this should be //+optional and omitempty.

yi0909: We use the ImageName to generate the VirtualMachineImage object name in the cluster, why would this field be optional?

janitha (approved):

api/v1alpha1/virtualmachineimage_types.go
bryanv: We should do so that these fields are Patch'able.

api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -95,6 +101,10 @@ type VirtualMachineImageStatus struct {
 	// Deprecated
 	PowerState string `json:"powerState,omitempty"`
 
+	// ImageName describes the display name of this VirtualMachineImage.
+	// optional
```
sreyasn, line 105: Nit:  I think the comment needs a '+'
`// +optional`

bryanv (approved):

aruneshpa (approved):

PR #78, opened 2021-06-30
Introduce CloudInitPrep type VirtualMachine Metadata transport

gandhipr:
This change introduces a const - VirtualMachineMetadataCloudInitPrepTransport = "CloudInitPrep" which will set the VirtualMachineMetadata ConfigMap data as cloud-init userdata.

### DIFF
--- a/api/v1alpha1/virtualmachine_types.go
+++ b/api/v1alpha1/virtualmachine_types.go
@@ -92,8 +92,8 @@ type VirtualMachineNetworkInterface struct {
 }
 
 // VirtualMachineMetadataTransport is used to indicate the transport used by VirtualMachineMetadata
-// Valid values are "ExtraConfig" and "OvfEnv".
-// +kubebuilder:validation:Enum=ExtraConfig;OvfEnv
+// Valid values are "ExtraConfig", "OvfEnv" and "CloudInit".
+// +kubebuilder:validation:Enum=ExtraConfig;OvfEnv;CloudInit
 type VirtualMachineMetadataTransport string
 
 const (
@@ -106,6 +106,16 @@ const (
 	// marked userConfigurable and already present in either OVF Properties of a VirtualMachineImage
 	// or as vApp properties on an existing VM or VMTX will be set, all others will be ignored.
 	VirtualMachineMetadataOvfEnvTransport VirtualMachineMetadataTransport = "OvfEnv"
+
+	// VirtualMachineMetadataCloudInitTransport indicates the data in
+	// the ConfigMap resource specified by VirtualMachineMetadata.ConfigMapName
+	// in the "user-data" key is cloud-init userdata.
+	//
+	// Please note that, despite the name, VirtualMachineMetadata has no
+	// relationship to cloud-init instance metadata.
+	//
+	// For more information, please refer to cloud-init's official documentation.
+	VirtualMachineMetadataCloudInitTransport VirtualMachineMetadataTransport = "CloudInit"
 )
 
 // VirtualMachineMetadata defines any metadata that should be passed to the VirtualMachine instance.  A typical use


aruneshpa:
Can you change the commit message and the PR subject to follow [this ](https://chris.beams.io/posts/git-commit/). Maybe change the commit message to something like:

```
Introduce CloudInitPrep type VirtualMachine Metadata transport
```

janitha (approved):

api/v1alpha1/virtualmachine_types.go
```diff
@@ -106,6 +106,10 @@ const (
 	// marked userConfigurable and already present in either OVF Properties of a VirtualMachineImage
 	// or as vApp properties on an existing VM or VMTX will be set, all others will be ignored.
 	VirtualMachineMetadataOvfEnvTransport VirtualMachineMetadataTransport = "OvfEnv"
+
+	// VirtualMachineMetadataCloudInitPrepTransport will set the VirtualMachineMetadata ConfigMap data as
```
akutz, line 110: This wording is _super_ confusing. Can we add a clarification here that it is explicitly _not_ **cloud-init** metadata.

akutz, line 110: Can we please rename this to:

```golang
VirtualMachineMetadataCloudInitTransport VirtualMachineMetadataTransport = "CloudInit"
```

We do not really need to reveal more than this is cloud-init IMO. In the future I would even like to rename the type to `VirtualMachineBootstrapRef` and let it point directly to an object reference, with that determining the provider. We need to move away from `ConfigMap` to `Secret` at some point to shield possibly sensitive data. For now let's just remove the specifics of the actual transport beyond saying `CloudInit`. Thanks!

akutz: Suggest:

```golang
// VirtualMachineMetadataCloudInitTransport indicates the data in
// the ConfigMap resource specified in
// VirtualMachineMetadata.ConfigMapName is cloud-init userdata.
//
// Please note, despite the name, that VirtualMachineMetadata has no
// relationship to cloud-init instance metadata. 
//
// For more information on cloud-init user data, please see
// https://cloudinit.readthedocs.io/en/latest/topics/format.html.
```

akutz (commented):
Looks great @gandhipr! I left two pieces of feedback. If you would respond to those and let me know if/how you plan to address them, I will happily mark this as approved!

api/v1alpha1/virtualmachine_types.go
janitha: Generally I'm in favor of making things generic as well, but in this case, being specific/explicit about the underlying transport is preferable imo. Also, we need some way to differentiate between CloudInitPrep vs. upcoming CloudInitNoCloud.

akutz: > Generally I'm in favor of making things generic as well, but in this case, being specific/explicit about the underlying transport is preferable imo. Also, we need some way to differentiate between CloudInitPrep vs. upcoming CloudInitNoCloud.

Is this important to the consumer of the API or the implementation? Is there any penalty for changing how the cloud-init data is transported?

gandhipr: I'll add this comment. This clarifies the confusion that could arise because of the name.

gandhipr: Even at the implementation level, this would help us in differentiating between the two.

akutz: > Even at the implementation level, this would help us in differentiating between the two.

When would we use two? Why would we need to do so? Also, more importantly, why would we expect someone interacting with the API to know which to choose?

akutz: Once we support NoCloud by dynamically building an ISO and mounting it to the VM, the need for any other datasource, such as Cloud Init Prep, is greatly diminished. The one exception _might_ be on VMC. Originally you could not use the NoCloud DS on VMC because you could not upload an ISO to a datastore on VMC and mount it to a VM. If this is still the case, we would likely need to use dual cloud-init transports.

However, I think if that need arises, that is a decision the VM Operator can and should make without requiring input from the user. For example, a flag that indicates VM Operator is deployed on VMC.

gandhipr: If specifying underlying transport to the customers is important then we need cloudInitPrep. Otherwise, how about having cloudInit as a generic name and if later we end up having both kinds, implementation details will remember that cloudInit means cloudInitPrep. But if we're switching to cloudInitNoCloud completely, then this won't be a problem.

akutz (approved):
Great job @gandhipr !!

api/v1alpha1/virtualmachine_types.go
```diff
@@ -106,6 +106,16 @@ const (
...
 	VirtualMachineMetadataOvfEnvTransport VirtualMachineMetadataTransport = "OvfEnv"
+
+	// VirtualMachineMetadataCloudInitTransport indicates the data in
+	// the ConfigMap resource specified by
+	// VirtualMachineMetadata.ConfigMapName is cloud-init userdata.
+	//
+	// Please note that, despite the name, VirtualMachineMetadata has no
+	// relationship to cloud-init instance metadata.
+	//
+	// For more information, please refer cloud-init's official document.
```
bryanv, line 117: document -> documentation, or just provide the link as in @akutz suggestion.

Do we say what the CM key is used for the data?

gandhipr: Even Andrew suggested not to the link because it is some .html link. This key is used for cloud-init userdata.

bryanv: Seems prudent though to document the _name_ of the expected key in the CM, because we are expecting a specific one.

gandhipr: Made changes as per our discussion.

PR #79, opened 2021-07-09
Rename Condition VSphereVMCustomization to GuestCustomization

janitha:
This change renames the `VSphereVMCustomization` condition to be something that is more generic, leaving room for supporting additional (non vsphere/vmtools) based customizations in the future.

### DIFF
--- a/api/v1alpha1/condition_consts.go
+++ b/api/v1alpha1/condition_consts.go
@@ -35,24 +35,26 @@ const (
 	// VirtualMachineImageNotFoundReason (Severity=Error) documents that the VirtualMachineImage specified in the VirtualMachineSpec
 	// is not available.
 	VirtualMachineImageNotFoundReason = "VirtualMachineImageNotFound"
+)
 
-	// VSphereVMCustomizationCondition exposes the status of vSphere VM Customization from within the guest OS, when available.
-	VSphereVMCustomizationCondition ConditionType = "VSphereVMCustomization"
+const (
+	// GuestCustomizationCondition exposes the status of guest customization from within the guest OS, when available.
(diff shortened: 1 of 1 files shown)


api/v1alpha1/condition_consts.go
```diff
@@ -35,24 +35,26 @@ const (
 	// VirtualMachineImageNotFoundReason (Severity=Error) documents that the VirtualMachineImage specified in the VirtualMachineSpec
 	// is not available.
 	VirtualMachineImageNotFoundReason = "VirtualMachineImageNotFound"
+)
 
-	// VSphereVMCustomizationCondition exposes the status of vSphere VM Customization from within the guest OS, when available.
-	VSphereVMCustomizationCondition ConditionType = "VSphereVMCustomization"
+const (
+	// GuestCustomizationCondition exposes the status of guest customization from within the guest OS, when available.
+	GuestCustomizationCondition ConditionType = "GuestCustomization"
```
aruneshpa, line 42: Is it possible that some of these Conditions are indeed specific to vSphere VM customization? In that case, making these generic does not add much value.

aruneshpa (approved):

api/v1alpha1/condition_consts.go
janitha: The { idle, pending, running, success, fail } are about as generic as you can get, and they do happen to align with what vsphere reports back via customizationStatus. Would you have suggestions to any additional ones? (or we can add them when they are relevant as well)

Also, if upstream cloud-init provides a way to query status natively, then these states would apply as-is. (I don't think there's any plan on cloud-init side, but this is something I'm looking in to opening on upstream cloud-init)

srajashekar-vmw (approved):

PR #80, opened 2021-07-27
Add Condition Types for VMTools running status

sreyasn:
This change adds condition types for the running status of
VMware tools, which will be exposed on the VirtualMachine CRD's status

### DIFF
--- a/api/v1alpha1/condition_consts.go
+++ b/api/v1alpha1/condition_consts.go
@@ -57,6 +57,17 @@ const (
 	GuestCustomizationFailedReason = "GuestCustomizationFailed"
 )
 
+const (
+	// VirtualMachineToolsCondition exposes the status of VMware Tools running in the guest OS, when available
+	VirtualMachineToolsCondition ConditionType = "VirtualMachineTools"
+
+	// VirtualMachineToolsNotRunningReason (Severity=Error) documents that VMware Tools is not running
+	VirtualMachineToolsNotRunningReason = "VirtualMachineToolsNotRunning"
+
+	// VirtualMachineToolsRunningReason (Severity=Info) documents that VMware Tools is running
+	VirtualMachineToolsRunningReason = "VirtualMachineToolsRunning"
+)
+
 // Common Condition.Reason used by VM Operator API objects.
 const (
 	// DeletingReason (Severity=Info) documents a condition not in Status=True because the underlying object it is currently being deleted.


api/v1alpha1/condition_consts.go
```diff
@@ -57,6 +57,20 @@ const (
...
+	// VirtualMachineToolsCondition exposes the status of VirtualMachine tools running in the guest OS, when available
+	VirtualMachineToolsCondition ConditionType = "VirtualMachineTools"
+
+	// VirtualMachineToolsNotRunningReason (Severity=Error) documents that VirtualMachine tools is not running
+	VirtualMachineToolsNotRunningReason = "VirtualMachineToolsNotRunning"
+
+	// VirtualMachineToolsRunningReason (Severity=Info) documents that VirtualMachine tools is running
+	VirtualMachineToolsRunningReason = "VirtualMachineToolsRunning"
+
+	// VirtualMachineToolsExecutingScriptsReason (Severity=Info) documents that VirtualMachine tools is starting
```
janitha, line 70: I would leave this state out (we don't need to mirror the vmodl api, and just having running/notrunning seems cleaner and sufficient)

janitha (approved):

akutz:
Should we have a condition type of "Unknown" or "NotDetected"?

janitha:
> Should we have a condition type of "Unknown" or "NotDetected"?

All conditions have a Unknown state (as part of corev1.ConditionStatus)

akutz:
Doh, duh. Sorry about that.

akutz (approved):
/lgtm

Thanks!

api/v1alpha1/condition_consts.go
```diff
@@ -57,6 +57,17 @@ const (
 	GuestCustomizationFailedReason = "GuestCustomizationFailed"
 )
 
+const (
+	// VirtualMachineToolsCondition exposes the status of VirtualMachine tools running in the guest OS, when available
```
bryanv, line 61: Isn't "VMware Tools" more commonly used, as in the product name and what the VC UI uses? I don't think I've ever seen this spelled out as "VirtualMachine tools"

sreyasn: yeah, VMware tools is more commonly used as the product name. I would like to use that in the comments too. Any restrictions from using "VMware" in our api docs?.

janitha: Should be fine to use that, after all, this is a product of VMware.

bryanv (approved):

PR #81, opened 2021-08-11
Introduce Instance Storage Volume APIs

jitinkumar2018:
This PR Intoduces Instance Storage Volumes APIs as a peer of the Virtual Devices spec. The Instance Storage struct contains the specification required to configure instance storage for virtual machines. An admin can create VM classes with instance storage volumes and assign them to Supervisor namespaces for consumption. These VM classes can be used to create virtual machines that have instance storage volumes attached as disks. The generated instance volumes are placed on vSAN Direct datastores of the underlying host.

### DIFF
--- a/api/v1alpha1/virtualmachineclass_types.go
+++ b/api/v1alpha1/virtualmachineclass_types.go
@@ -21,6 +21,28 @@ type DynamicDirectPathIODevice struct {
 	CustomLabel string `json:"customLabel,omitempty"`
 }
 
+// InstanceStorage provides information used to configure instance
+// storage volumes for a VirtualMachine.
+type InstanceStorage struct {
+	// StorageClass refers to the name of a StorageClass resource used to
+	// provide the storage for the configured instance storage volumes.
+	// The value of this field has no relationship to or bearing on the field
+	// virtualMachine.spec.storageClass. Please note the referred StorageClass
+	// must be available in the same namespace as the VirtualMachineClass that
+	// uses it for configuring instance storage.
+	StorageClass string `json:"storageClass,omitempty"`
+
+	// Volumes describes the number and size of instance storage volumes
+	// created for a VirtualMachine that uses this VirtualMachineClass.
+	Volumes []InstanceStorageVolume `json:"volumes"`
+}
+
+// InstanceStorageVolume contains information required to create an
+// instance storage volume on a VirtualMachine.
+type InstanceStorageVolume struct {
+	Size resource.Quantity `json:"size"`
+}
+
 // VirtualDevices contains information about the virtual devices associated with a VirtualMachineClass.
 type VirtualDevices struct {
 	// +optional
@@ -35,6 +57,8 @@ type VirtualMachineClassHardware struct {
 	Memory  resource.Quantity `json:"memory,omitempty"`
 	// +optional
 	Devices VirtualDevices    `json:"devices,omitempty"`
+	// +optional
+	InstanceStorage InstanceStorage `json:"instanceStorage,omitempty"`
 }
 
 // VirtualMachineResourceSpec describes a virtual hardware policy specification.


api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -21,6 +21,31 @@ type DynamicDirectPathIODevice struct {
...
+// InstanceStorageVolumeSpec describes an instance storage volume specification.
+type InstanceStorageVolumeSpec struct {
+	Resources InstanceStorageVolumeResource `json:"resources"`
+}
+
+// InstanceStorageVolumeResource describes an instance storage volume resource specification.
+type InstanceStorageVolumeResource struct {
+	Requests resource.Quantity `json:"requests"`
+}
+
```
janitha, line 48: `InstanceStorageVolumeSpec` and `InstanceStorageVolumeResource` seems verbose and unnecessary. 

Can we omit those and just have: 
```
type InstanceStorageVolume struct {
  Requests resource.Quantity `json:"requests"`
}
```

api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -21,6 +21,31 @@ type DynamicDirectPathIODevice struct {
 	CustomLabel string `json:"customLabel,omitempty"`
 }
 
+// InstanceStorage encapsulates information about instance storage.
+type InstanceStorage struct {
+	// StorageClass describes the configuration of instance storage related attributes on the VirtualMachine instance.
```
janitha, line 26: Could you expand the description to state what storageClass will be used if this field is omitted (I assume it's the per-namespace assigned one?)

akutz: For example:

```
StorageClass refers to the name of a StorageClass resource used to
provide the storage for the configured, instance storage volumes.

The value of this field has no relationship to or bearing on the field
virtualMachine.spec.storageClass.

Please note the referred StorageClass must be available in the same
namespace as the VirtualMachineClass that uses it for configuring
instance storage.
```

api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -21,6 +21,31 @@ type DynamicDirectPathIODevice struct {
...
+	// +optional
+	StorageClass string `json:"storageClass,omitempty"`
+
+	// Volumes describes a list of InstanceStorageVolumes that are to be configured on the VirtualMachine.
+	Volumes []InstanceStorageVolume `json:"instanceStorage"`
```
akutz, line 31: `s/json:"instanceStorage"/json:"volumes"/`

api/v1alpha1/virtualmachineclass_types.go
akutz: Hard agree. Non CRDs do not need to follow the spec pattern. Please refactor to what @janitha recommends above.

api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -21,6 +21,31 @@ type DynamicDirectPathIODevice struct {
 	CustomLabel string `json:"customLabel,omitempty"`
 }
 
+// InstanceStorage encapsulates information about instance storage.
+type InstanceStorage struct {
+	// StorageClass describes the configuration of instance storage related attributes on the VirtualMachine instance.
+	// +optional
+	StorageClass string `json:"storageClass,omitempty"`
```
akutz, line 28: This field should not be optional.

akutz (changes requested):

api/v1alpha1/virtualmachineclass_types.go
bryanv: IMO `Size` would much more descriptive field name

akutz: In addition, please indicate with a comment the unit of size.

api/v1alpha1/virtualmachineclass_types.go
akutz: Yeah, I'm aware, but I also don't know how to best make the distinction since VM Classes *are* bound to specific namespaces. If that makes sense.

api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -21,6 +21,30 @@ type DynamicDirectPathIODevice struct {
 	CustomLabel string `json:"customLabel,omitempty"`
 }
 
+// InstanceStorage encapsulates information about instance storage.
+// StorageClass refers to the name of a StorageClass resource used to
```
akutz, line 25: Please leave empty lines between new topic, ex.

```go
// InstanceStorage provides information used to configure instance
// storage volumes for a VM.
//
// StorageClass refers to the name of a StorageClass resource used to
// provide the storage for the configured, instance storage volumes.
//
// The value of this field has no relationship to or bearing on the field
// virtualMachine.spec.storageClass.
//
// Please note the referred StorageClass must be available in the same
// namespace as the VirtualMachineClass that uses it for configuring
// instance storage.
```

api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -21,6 +21,30 @@ type DynamicDirectPathIODevice struct {
...
+// provide the storage for the configured, instance storage volumes.
+// The value of this field has no relationship to or bearing on the field
+// virtualMachine.spec.storageClass.
+// Please note the referred StorageClass must be available in the same
+// namespace as the VirtualMachineClass that uses it for configuring
+// instance storage.
+type InstanceStorage struct {
+	// StorageClass describes the configuration of instance storage related
```
akutz, line 33: Maybe...

```go
// StorageClass refers to the name of the StorageClass resource used to
// provide the storage for the configured instance storage volumes.
```

api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -21,6 +21,30 @@ type DynamicDirectPathIODevice struct {
...
+	// attributes on the VirtualMachine instance.
+	StorageClass string `json:"storageClass,omitempty"`
+
+	// Volumes describes a list of InstanceStorageVolumes that are to be
+	// configured on the VirtualMachine. 
+	Volumes []InstanceStorageVolume `json:"volumes"`
```
akutz, line 39: Maybe...

```go
// Volumes describes the number and size of instance storage volumes
// created for a VM that uses this VM Class.
```

api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -21,6 +21,30 @@ type DynamicDirectPathIODevice struct {
...
+}
+
+// InstanceStorageVolume contains configuration corresponding to an
```
akutz, line 42: Maybe..

```go
// InstanceStorageVolume contains information required to create an
// instance storage volume on a VM.
```

api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -21,6 +21,30 @@ type DynamicDirectPathIODevice struct {
...
+// instance storage volume.
+type InstanceStorageVolume struct {
+	Size resource.Quantity `json:"size"`
```
akutz, line 45: We're using `resource.Quantity`, which can be specified in multiple units. Is there some minimum size we should document here? Or is that unimportant since people won't be creating these directly anyway?

janitha (approved):

api/v1alpha1/virtualmachineclass_types.go
narendramadanapalli: Size is a required parameter that is not directly specified by the user of VM class. I think the minimum size depends on the type (small, xlarge etc) of VM class.

akutz: Fair.

api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -21,6 +21,34 @@ type DynamicDirectPathIODevice struct {
 	CustomLabel string `json:"customLabel,omitempty"`
 }
 
+// InstanceStorage provides information used to configure instance
+// storage volumes for a VM.
+//
+// StorageClass refers to the name of a StorageClass resource used to
+// provide the storage for the configured instance storage volumes.
+//
+// The value of this field has no relationship to or bearing on the field
```
bryanv, line 30: "this field" is confusing in a separate paragraph

api/v1alpha1/virtualmachineclass_types.go
bryanv: I think @akutz suggestion is much clearer should be used for the field description instead of the putting it above in the struct description.

api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -21,6 +21,34 @@ type DynamicDirectPathIODevice struct {
...
+//
+// Please note the referred StorageClass must be available in the same
+// namespace as the VirtualMachineClass that uses it for configuring
+// instance storage.
+type InstanceStorage struct {
+	// StorageClass describes the configuration of instance storage related
+	// attributes on the VirtualMachine instance.
+	StorageClass string `json:"storageClass,omitempty"`
+
+	// Volumes describes the number and size of instance storage volumes
```
bryanv, line 41: Suggestion:

> // Volumes describes the instance storage volumes created for a VirtualMachine that uses the VirtualMachineClass.

A general note for elsewhere in the MR: we've typically avoided using abbreviations in the docs to be more precise about what is being referred to

akutz (approved):
LGTM!

api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -35,6 +57,8 @@ type VirtualMachineClassHardware struct {
 	Memory  resource.Quantity `json:"memory,omitempty"`
 	// +optional
 	Devices VirtualDevices    `json:"devices,omitempty"`
+	// +optional
+	InstanceStorage InstanceStorage `json:"instanceStorage,omitempty"`
```
akutz, line 61: We agreed that InstanceStorage should be a peer of Devices, not part of it. Please check the internal VMODL change and align with that. Ping Narendra privately for a link to that review. Thanks!

akutz (changes requested):

api/v1alpha1/virtualmachineclass_types.go
akutz: My mistake. The GH view on my phone confused me. The current model is aligned with what I was thinking.image

akutz (approved):

PR #82, opened 2021-08-11
Add a DeployedCondition for the VirtualMachine object

sreyasn:
This change adds a conditionType for the deployment status of a virtualmachine on
the infrastructure provider.

### DIFF
--- a/api/v1alpha1/condition_consts.go
+++ b/api/v1alpha1/condition_consts.go
@@ -38,6 +38,14 @@ const (
 )
 
 const (
+	// DeployedCondition exposes the deployment status of a VirtualMachine on the infrastructure provider.
+	DeployedCondition ConditionType = "Deployed"
+
+	// DeployFailedReason (Severity=Error) documents that the deployment of the VirtualMachine was not successful.
+	DeployFailedReason = "DeployFailed"
+)
+
+const (
 	// GuestCustomizationCondition exposes the status of guest customization from within the guest OS, when available.
 	GuestCustomizationCondition ConditionType = "GuestCustomization"
 


janitha (approved):

api/v1alpha1/condition_consts.go
```diff
@@ -37,6 +37,14 @@ const (
 	VirtualMachineImageNotFoundReason = "VirtualMachineImageNotFound"
 )
 
+const (
+	// DeployedCondition exposes the status of the VirtualMachine deployment.
```
aruneshpa, line 41: I think the `VirtualMachine deployment` reads as `VirtualMachineDeployment`. Could you update this to reflect that we are talking about the VM being deployed on the infrastructure provider?

aruneshpa (approved):

api/v1alpha1/condition_consts.go
sreyasn: will do. Thanks for taking a look

api/v1alpha1/condition_consts.go
```diff
@@ -37,6 +37,14 @@ const (
 	VirtualMachineImageNotFoundReason = "VirtualMachineImageNotFound"
 )
 
+const (
+	// DeployedCondition exposes the deployment status of a VirtualMachine on the infrastructure provider.
+	DeployedCondition ConditionType = "Deployed"
```
akutz, line 42: I'm not sure I understand what this means. The VM hardware deployed successfully? The virtual hardware failed? The guest customization succeeded / failed? Both? I am not sure this makes sense as-is. Let's revisit this a bit more. I don't know that aggregate conditions make sense, and we already have a guest customization condition, so is deploy separate or a superset?

api/v1alpha1/condition_consts.go
```diff
@@ -37,6 +37,14 @@ const (
...
+
+	// DeployFailedReason (Severity=Error) documents that the deployment of the VirtualMachine was not successful.
+	DeployFailedReason = "DeployFailed"
```
akutz, line 45: Same as above.

akutz (changes requested):

api/v1alpha1/condition_consts.go
sreyasn: Deploy is separate, It's synonymous the Created/Create in VM operator. 

guest customization is triggered in our Update path after Create/Deploy passes.

akutz: > Deploy is separate, It's synonymous to Created/Create in VM operator.

The issue is that "deploy" is vague enough to be taken multiple ways. Plus, [conditions exist](https://maelvls.dev/kubernetes-conditions/) as a way to be more distinct than allowed by a top-level status field such as `phase`. So let's come up with a better word than `deployed` as it seems to generic and could be mistaken to mean guest customization as well.

PR #83, opened 2021-08-11
Add Cluster to the ClusterModuleStatus

bryanv:
This field will contain the domain-XXX that corresponds to the
ClusterComputeResource. The VC Cluster Modules API is not very
friendly for our needs: the module create API does not take a
name or other identifier, so we're force to keep the module ID
in our status to be able to look it up. With future HA support,
we need create the modules on each cluster, so we also need the
cluster in the status.

### DIFF
--- a/api/v1alpha1/virtualmachinesetresourcepolicy_types.go
+++ b/api/v1alpha1/virtualmachinesetresourcepolicy_types.go
@@ -49,8 +49,9 @@ type VirtualMachineSetResourcePolicyStatus struct {
 }
 
 type ClusterModuleStatus struct {
-	GroupName  string `json:"groupname"`
-	ModuleUuid string `json:"moduleUUID"`
+	GroupName   string `json:"groupname"`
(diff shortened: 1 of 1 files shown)


api/v1alpha1/virtualmachinesetresourcepolicy_types.go
```diff
@@ -51,6 +51,7 @@ type VirtualMachineSetResourcePolicyStatus struct {
 type ClusterModuleStatus struct {
 	GroupName  string `json:"groupname"`
 	ModuleUuid string `json:"moduleUUID"`
+	Cluster    string `json:"cluster"`
```
akutz, line 54: Please rename this to `ClusterMoId` and `clusterMoId` in the JSON tag. Because the example value provided is the managed object ID for a cluster.

janitha (approved):

aruneshpa (approved):
LGTM.

PR #84, opened 2021-08-19
Quick fix to make instanceStorage.volumes omitempty

bryanv:
Otherwise existing classes are invalid:

 ... is invalid: spec.hardware.instanceStorage.volumes: Invalid value: \"null\":
     spec.hardware.instanceStorage.volumes in body must be of type array: \"null\"",

Tweak field wording a bit.

This should use later use CRD validation to enforce a non-empty array.

The optional structs of the VirtualMachineClassHardware probably made/make
more sense as pointers.


aruneshpa (approved):
LGTM. Thanks for the quick fix.

janitha (approved):

PR #86, opened 2021-09-22
Add 'Description' field to VM Class CRD

zyiyi11:
For [VMSVC-307](https://jira.eng.vmware.com/browse/VMSVC-307)
VI admins can provide a description of a VM class when they create them from the VC UI (or VAPI). However, that description is not propagated to the custom resource since the CRD doesn't support it. 
To expose description field on -o yaml/-o json, I added a new type VirtualMachineClassDescription and include it in VirtualMachineClass.

### DIFF
--- a/api/v1alpha1/virtualmachineclass_types.go
+++ b/api/v1alpha1/virtualmachineclass_types.go
@@ -90,6 +90,10 @@ type VirtualMachineClassSpec struct {
 	// policy.  The configuration specified in this field is used to customize various policies related to
 	// infrastructure resource consumption.
 	Policies VirtualMachineClassPolicies `json:"policies,omitempty"`
+
+	// Description describes the configuration of the VirtualMachineClass which is not related to virtual hardware
+	// or infrastructure policy. This field is used to address remaining specs about this VirtualMachineClass.
+	Description VirtualMachineClassDescription `json:"description,omitempty"`
 }
 
 // VirtualMachineClassStatus defines the observed state of VirtualMachineClass.  VirtualMachineClasses are immutable,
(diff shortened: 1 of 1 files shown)


api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -118,6 +122,7 @@ type VirtualMachineClass struct {
 
 	Spec   VirtualMachineClassSpec   `json:"spec,omitempty"`
 	Status VirtualMachineClassStatus `json:"status,omitempty"`
+	Description VirtualMachineClassDescription `json:"description,omitempty"`
```
janitha, line 125: The CRDs root shouldn't include anything else besides the `spec` and [status](url). I think the Description is better suited to be in the Spec (since it's meant to be set by the VirtualMachineClass creator)

api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -90,13 +90,21 @@ type VirtualMachineClassSpec struct {
 	// policy.  The configuration specified in this field is used to customize various policies related to
 	// infrastructure resource consumption.
 	Policies VirtualMachineClassPolicies `json:"policies,omitempty"`
+
+	// Description describes the configuration of the VirtualMachineClass which is not related to virtual hardware
+	// or infrastructure policy. This field is used to address remaining specs about this VirtualMachineClass.
+	Description VirtualMachineClassDescription `json:"description,omitempty"`
```
janitha, line 96: This can be a simple string (rather than a struct VirtualMachineClassDescription). This maps 1:1 with the vapi vmclass description, which is also a string.

janitha (approved):

PR #91, opened 2021-12-02
Introduce annotation to pause a VM reconcile

aruneshpa:
In order to support VADP based backup/restore of VM Service VMs, we
need to pause the reconcile while the VM is recreated from backup on
the infrastructure. This change introduces a pause annotation to
achieve this. When set, VM operator will not reconcile this VM until
this annotation is removed.

### DIFF
--- a/api/v1alpha1/virtualmachine_types.go
+++ b/api/v1alpha1/virtualmachine_types.go
@@ -43,6 +43,14 @@ const (
 	Unknown VMStatusPhase = "Unknown"
 )
 
+// PauseAnnotation is an annotation that can be applied to any VirtualMachine object to prevent VM Operator from
+// reconciling the object with the vSphere infrastructure.  VM Operator checks the presence of this annotation to
+// skip the reconcile of a VirtualMachine.
+//
+// This can be used when a Virtual Machine needs to be modified out-of-band of VM Operator on the infrastructure
+// directly (e.g., during a VADP based Restore operation).
+const PauseAnnotation = GroupName + "/pause-reconcile"
+
 // VirtualMachinePort is unused and can be considered deprecated.
 type VirtualMachinePort struct {
 	Port     int             `json:"port"`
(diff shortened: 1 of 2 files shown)


api/v1alpha1/virtualmachine_types.go
```diff
@@ -43,6 +43,11 @@ const (
 	Unknown VMStatusPhase = "Unknown"
 )
 
+// Annotation label to pause the reconciliation of a VM.  If set, VM operator will not reconcile a VirtualMachine
+// resource.  This can be used when a Virtual Machine needs to be modified out-of-band of VM operator on the
+// infrastructure directly (e.g., during a VADP based Restore operation).
+const PauseReconcileAnnotationLabel = GroupName + "/pause-reconcile"
```
janitha, line 49: What should the label value be? You could mention it in the comment, or introduce a variable/s for possible values.

janitha (approved):

api/v1alpha1/virtualmachine_types.go
bryanv: "AnnotationLabel" is very confusing. Seems this is used as an annotation so I'd suggest to drop the Label part.

aruneshpa: Done

bryanv (approved):

api/v1alpha1/virtualmachine_types.go
aruneshpa: > What is the expected label value? You could mention it in the comment, or introduce a variable/s for possible values.

The value of this annotation does not matter. We will only check the presence of the annotation to detect if we need to skip the reconcile.
Ref: CAPI does something similar: 
- https://github.com/kubernetes-sigs/cluster-api/blob/3da7a18b40810cd8a33975e81c89978c6c8beedc/api/v1beta1/common_types.go#L61No
- https://github.com/kubernetes-sigs/cluster-api/blob/58905fb5140a7844c3e24aa70d255fb8f9d0c491/util/annotations/helpers.go#L41

PR #93, opened 2022-04-06
Introduce VirtualMachineConfigSpec

sreyasn:
This changes introduces the VirtualMachineConfigSpec type to support additional virtual machine configuration settings for VMs created from a VirtualMachineClass


api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -8,6 +8,30 @@ import (
...
+	// serialized to XML and base64-encoded.
+	ConfigSpecProviderXml VirtualMachineConfigSpecProvider = "XML"
+)
+
+// VirtualMachineConfigSpec contains additional virtual machine
+// configuration settings including hardware specifications for the VirtualMachine
+type VirtualMachineConfigSpec struct {
+	// Provider indicates which field in VirtualMachineConfigSpec provides the additional virtual machine
+	// configuration settings.
+	Provider VirtualMachineConfigSpecProvider `json:"provider"`
```
bryanv, line 27: Is this something that is really needed? Say like later we add `JSON` or `TemplateMoID` fields, would multiple fields be set that we need this to select which one to use?

Or can the validation webhook assert that only one of these field(s) is set?

sreyasn: good point. We might end up having a `JSON` field but I don't see how multiple fields can be set unless someone manually tries to a VM Class (ie) wcpsvc should do the right thing. Even if it does happen, our validation webhook could handle the case so that only one field is set at a time.

akutz: So there's precedent with MetadataTransportType, which is why I elected to define an explicit Provider (type). There are no unions in Go, so without an explicit type, we don't know for sure. I suppose I'm fine with something implicitly as long as the order of checking is explicit as well.

Because I'm of the mind it's easier to add than remove when it comes to CRDs, I'm fine omitting the provider type for now as long as we add a note to the `VirtualMachineConfigSpec` type that the `XML` field may be the only value for now, but in the future that may change.

sreyasn: sounds good.

bryanv: IDK if VirtualMachineMetadataTransport is quite analogous here because we really can't infer the desired transport type from the other fields set - the data for various types can come from either a ConfigMap or Secret. 

Here I think we can, and it is one less thing to enter and validate when it seems like the choice is already implicit from other validations we want on this struct (one and only one "data" field will be non-empty).

janitha (approved):

api/v1alpha1/virtualmachineclass_types.go
```diff
@@ -8,6 +8,16 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// VirtualMachineConfigSpec contains additional virtual machine
+// configuration settings including hardware specifications for the VirtualMachine
+type VirtualMachineConfigSpec struct {
+	// XML contains a vim.vm.ConfigSpec object that has been serialized to XML
+	// and base64-encoded. We only support XML for now, but that may change
+	// in the future.
```
bryanv, line 16: > We only support XML for now, but that may change in the future.

I think this makes more sense in the description for this struct, or omit it since XML is the only field defined.

akutz (approved):

PR #94, opened 2022-05-17
Add a NoDefaultNic Annotation.

yi0909:
When the no-network annotation exists, no default network will be applied to the VM.
Otherwise, vm-operator will add a default network to the VM based on the network provider.

### DIFF
--- a/api/v1alpha1/virtualmachine_types.go
+++ b/api/v1alpha1/virtualmachine_types.go
@@ -43,13 +43,25 @@ const (
 	Unknown VMStatusPhase = "Unknown"
 )
 
-// PauseAnnotation is an annotation that can be applied to any VirtualMachine object to prevent VM Operator from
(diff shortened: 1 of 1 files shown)


akutz:
I want to drop a hold on this and think on it over the weekend. I'm the one who suggested it, and I feel it is misleading as it is. I'll circle back to this on Monday.

akutz:
Why is the pause annotation introduced as part of this PR?

yi0909:
This PR doesn't introduce the pause annotation, but just change its place. I think they are all annotations so I put them into one `const` block. @akutz

yi0909:
This PR doesn't introduce the pause annotation, but just change its place. I think they are all annotations so I put them into one `const` block. @akutz

akutz:
Can we please update the docs for the new annotation to indicate when this will prevent what behavior? What if an existing VM has no NIC? When does the new behavior get applied?

PR #97, opened 2022-07-18
Support for GuestOS ID / Family

akutz:
This patch adds support for specifying the guest operating system's ID and family.


api/v1alpha1/guestosdescriptor_types.go
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//go:generate go run ./internal/generate_guestosidentifiers.go
+
+package v1alpha1
+
+// VirtualMachineGuestOSDescriptor is used to specify information about the
```
aruneshpa, line 8: Accidental deletion of text?

api/v1alpha1/guestosdescriptor_types.go
```diff
@@ -0,0 +1,58 @@
...
+	// +optional
+	Family VirtualMachineGuestOSFamily `json:"family,omitempty"`
+}
+
+// VirtualMachineGuestOSFamily specifies the family of the guest OS.
+type VirtualMachineGuestOSFamily string
+
+const (
+	VirtualMachineGuestOSFamilyWindows       VirtualMachineGuestOSFamily = "windowsGuest"
+	VirtualMachineGuestOSFamilyLinux                                     = "linuxGuest"
```
aruneshpa, line 33: Do we also need to specify `VirtualMachineGuestOSFamily` type for all these consts, or is this Golang magic that automatically assigns them the types? Either way, I would vote for being explicit in the name of readability. (Same for Guest OS Identifiers)

aruneshpa: :-D 
I actually tried this in playground by checking the type of two consts, before leaving the comment.  I still thought you knew something that I didn't about Golang internals (as you often do).

PR #99, opened 2022-08-10
VirtualMachinePublishRequest API

akutz:
This patch introduces the new VirtualMachinePublishRequest API for supporting self-service publication of VM Service VMs as new VM Images in the same namespace as the VM being published.

Please note this patch also updates the Kubernetes dependencies to v0.23.1 in order to leverage the metav1.Condition type from the Kubernetes API machinery project.

The Go module version of Go is also updated to v1.17 since v1.13 is getting rather long in the tooth.


api/v1alpha1/virtualmachinepublishrequest_types.go
```diff
@@ -0,0 +1,530 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// VirtualMachinePublishRequestConditionSourceValid is the Type for a
+	// VirtualMachinePublishRequest resource's status condition.
+	//
+	// The condition's status is set to true only when the information
+	// that describes the source side of the publication has been validated.
+	VirtualMachinePublishRequestConditionSourceValid = "SourceValid"
```
yi0909, line 18: Maybe we can move these conditions to https://github.com/vmware-tanzu/vm-operator-api/blob/master/api/v1alpha1/condition_consts.go?

api/v1alpha1/virtualmachinepublishrequest_types.go
```diff
@@ -0,0 +1,530 @@
...
+) {
+	vmpr.markCondition(
+		VirtualMachinePublishRequestConditionUploaded,
+		status,
+		args...,
+	)
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Namespace,shortName=vmpub
```
yi0909, line 501: Typo: Namespaced?

api/v1alpha1/virtualmachinepublishrequest_types.go
```diff
@@ -0,0 +1,530 @@
...
+	//   * Complete
+	//
+	// +optional
+	Ready bool `json:"ready,omitempty"`
+
+	// Conditions is a list of the latest, available observations of the
+	// request's current state.
+	//
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
yi0909, line 240: vm-operator-api has its own Condition type https://github.com/vmware-tanzu/vm-operator-api/blob/master/api/v1alpha1/condition_types.go#L45. Maybe we can use this?

api/v1alpha1/virtualmachinepublishrequest_types.go
```diff
@@ -0,0 +1,530 @@
...
+// applying a VirtualMachinePublishRequest resource that has the same name
+// as said VM in the same namespace as said VM.
+type VirtualMachinePublishRequestSpec struct {
+	// Source is the source of the publication request, ex. a VirtualMachine
+	// resource.
+	//
+	// If this value is omitted then the publication controller checks to
+	// see if there is a resource with the same name as this
+	// VirtualMachinePublishRequest resource, an API version equal to
+	// spec.source.apiVersion, and a kind equal to spec.source.kind. If such
```
yi0909, line 152: If the source is omitted, how could we get spec.source.apiVersion and spec.source.kind?

api/v1alpha1/virtualmachinepublishrequest_types.go
```diff
@@ -0,0 +1,530 @@
...
+	Source VirtualMachinePublishRequestSource `json:"source,omitempty"`
+
+	// Target is the target of the publication request, ex. item
+	// information and a ContentLibrary resource.
+	//
+	// If this value is omitted, the controller uses spec.source.name + "-image"
+	// as the name of the published item. Addtionally, when omitted the
+	// controller attempts to identify the target location by matching a
+	// resource with an API version equal to spec.target.location.apiVersion, a
+	// kind equal to spec.target.location.kind, w/ the label
```
yi0909, line 165: Same as above, how could we get spec.target.location.kind/apiVersion if the entire target value is omitted?

api/v1alpha1/virtualmachinepublishrequest_types.go
akutz: Could, but these are specific to this type and make more sense here IMO.

api/v1alpha1/virtualmachinepublishrequest_types.go
akutz: Good catch!

api/v1alpha1/virtualmachinepublishrequest_types.go
akutz: I'd like to move us to using the standard `metav1.Condition` type and improve on that one where necessary.

api/v1alpha1/virtualmachinepublishrequest_types.go
akutz: Because this not a pointer and thus will be subject to defaulting, and thus you should look at the default values for the fields in the type `VirtualMachinePublishRequestSource`

api/v1alpha1/virtualmachinepublishrequest_types.go
akutz: Same as the above answer but look in `VirtualMachinePublishRequestTarget`

api/v1alpha1/virtualmachinepublishrequest_types.go
```diff
@@ -0,0 +1,530 @@
...
+		message = args[1]
+	}
+	if reason == "" && status == metav1.ConditionTrue {
+		reason = "success"
+	} else {
+		reason = string(status)
+	}
+
+	c := vmpr.getCondition(conditionType)
+	if c != nil && c.Status != status {
```
bryanv, line 340: This looks like a potential usage issue: if the Condition's Status is the _same_ we shouldn't add another entry to the Status.Conditions[]s (like somebody just calls MarkComplete() at the end of the reconcile). The CAPI-inspired Conditions that we use elsewhere will update the existing condition with the new status, reason, message (with a little bit of smarts around when it either updates or preserves the LastTransitionTime).

api/v1alpha1/virtualmachinepublishrequest_types.go
```diff
@@ -0,0 +1,530 @@
...
+	// kind equal to spec.target.location.kind, w/ the label
+	// "imageregistry.vmware.com/default".
+	//
+	// Please note that while optional, if a VirtualMachinePublishRequest sans
+	// target information is applied to a namespace without a default
+	// publication target, then the VirtualMachinePublishRequest resource
+	// will be marked in error.
+	//
+	// +optional
+	Target VirtualMachinePublishRequestTarget `json:"target,omitempty"`
```
dilyar85, line 174: This might implementation detail of the controller, but I wonder if we could utilize webhooks for checking here (e.g. mutation webhook to apply the default publication target and validating webhook to deny the request if no default exists)? I think it's more k8s native and also returns errors faster and easier back to users (i.e. in the kubectl `apply/create` command rather than in the resource's condition, which requires a kubectl `describe` or  `get -oyaml` .

api/v1alpha1/virtualmachinepublishrequest_types.go
```diff
@@ -0,0 +1,530 @@
...
+	// resource will be allowed to exist once the publication operation
+	// completes. After the TTL expires, the resource will be automatically
+	// deleted without the user having to take any direct action.
+	//
+	// If this field is unset then the request resource will not be
+	// automatically deleted. If this field is set to zero then the request
+	// resource is eligible for deletion immediately after it finishes.
+	//
+	// +optional
+	TTLSecondsAfterFinished *int64 `json:"ttlSecondsAfterFinished,omitempty"`
```
dilyar85, line 186: I wonder if K8s would set the value to 0 if it's not provided and therefore cause the unexpected deletion immediately?

dilyar85: Oh nvm! I saw it's a pointer type now.

api/v1alpha1/virtualmachinepublishrequest_types.go
```diff
@@ -0,0 +1,530 @@
...
+	// request after the publication operation completes.
+	//
+	// This field will not be set until the VirtualMachineImage resource
+	// is realized.
+	//
+	// +optional
+	ImageName string `json:"imageName,omitempty"`
+
+	// Ready is set to true only when the VM has been published successfully
+	// and the new VirtualMachineImage resource is ready.
```
dilyar85, line 221: Do we have (or plan to add) any readiness checks on our VirtualMachineImage resource? If not, feel like "exist" or "realized" sounds more accurate than "ready".

api/v1alpha1/virtualmachinepublishrequest_types.go
```diff
@@ -0,0 +1,530 @@
...
+}
+
+// IsSourceValid returns true if there is a status condition of Type=SourceValid
+// and Status=True.
+func (vmpr VirtualMachinePublishRequest) IsSourceValid() bool {
+	c := vmpr.getCondition(VirtualMachinePublishRequestConditionSourceValid)
+	if c != nil && c.Status == metav1.ConditionTrue {
+		return true
+	}
+	return false
```
dilyar85, line 264: Perhaps we can simply have `return c != nil && c.Status == metav1.ConditionTrue` here and below as well?

api/v1alpha1/virtualmachinepublishrequest_types.go
```diff
@@ -0,0 +1,530 @@
...
+		reason  string
+	)
+	if len(args) > 0 {
+		reason = args[0]
+	}
+	if len(args) > 1 {
+		message = args[1]
+	}
+	if reason == "" && status == metav1.ConditionTrue {
+		reason = "success"
```
dilyar85, line 334: I might be fussy here but looks like the value of `Reason` should be a CamelCase string, according to https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition

api/v1alpha1/virtualmachinepublishrequest_types.go
```diff
@@ -0,0 +1,530 @@
...
+// the current time in UTC. Please note this field is updated every
+// time this function is called. If this function is called with a
+// status of anything other than True, the value of status.completionTIme
+// is zeroed.
+//
+// This function also sets status.ready=true IFF the provided status
+// is True and all other conditions present in the resource have a
+// status=True.
+//
+// The variadic paramter args may be used to set the condition's
```
dilyar85, line 372: "parameter" looks like it's copy-pasted as it appears in multiple places below too :)

dilyar85 (approved):
Left a few comments but overall the PR LGTM. Thanks Andrew for adding this!

api/v1alpha1/virtualmachinepublishrequest_types.go
dilyar85: One more thing to clarify here, by saying "exist once the publication operation _completes_", does it mean the auto-deletion only occurs when the publish request completes successfully (i.e. after reaching its "Complete" condition)?

api/v1alpha1/virtualmachinepublishrequest_types.go
akutz: Sure, you could, but that's an impl detail.

api/v1alpha1/virtualmachinepublishrequest_types.go
akutz: Completes doesn't connote success or fail. Ready is only true if all conditions are met. If the publication ever fails terminally then complete is true, regardless of success.

api/v1alpha1/virtualmachinepublishrequest_types.go
akutz: Ready is a standard status term. It means the VM Image was published and ready for use.

go.mod
```diff
@@ -3,8 +3,8 @@ module github.com/vmware-tanzu/vm-operator-api
 go 1.17
 
 require (
-	k8s.io/api v0.23.5
-	k8s.io/apimachinery v0.23.5
+	k8s.io/api v0.23.1
```
yi0909, line 6: Is there a reason why we downgrade the version here?

yi0909:
Maybe we can add a Condition indicating that publish work is in progress? So when we reconcile the VMPub object and the CreateOvf hasn't finished, we won't trigger another CreateOVF request.

api/v1alpha1/virtualmachinepublishrequest_types.go
akutz: I agree with @yi0909 that we will continue to use the existing Condition type for now.

akutz:
> Maybe we can add a Condition indicating that publish work is in progress? So when we reconcile the VMPub object and the CreateOvf hasn't finished, we won't trigger another CreateOVF request.

We discussed this offline and may introduce this at a later date once we decide how to treat polarity and conditions.

PR #101, opened 2022-08-19
add vAppConfig as a new VirtualMachineMetadataTransport

zyiyi11:
This change adda s new `VirtualMachineMetadataTransport` called `vAppConfig`. 
This will set vAppConfig properties for VMs imported through OVFs. It will allow images bootstrap themselves with OVF properties that define the network information instead of relying on Guest OS Customization (GOSC).


api/v1alpha1/virtualmachine_types.go
```diff
@@ -129,6 +129,12 @@ const (
 	// or as vApp properties on an existing VM or VMTX will be set, all others will be ignored.
 	VirtualMachineMetadataOvfEnvTransport VirtualMachineMetadataTransport = "OvfEnv"
 
+	// VirtualMachineMetadataVAppConfigTransport indicates that the data set in the VirtualMachineMetadata Transport Resource,
+	// i.e., a ConfigMap or Secret, will be vApp properties on the VM, which will be exposed as vAppConfig to the Guest VM. Only properties
+	// marked userConfigurable and already present in either OVF Properties of a VirtualMachineImage
+	// or as vApp properties on an existing VM or VMTX will be set, all others will be ignored.
+	VirtualMachineMetadataVAppConfigTransport VirtualMachineMetadataTransport = "vAppConfig"
```
akutz, line 136: A few things:

* Please try not to exceed 80 columns for comments
* Please add a note that indicates this transport does not use Guest OS customization for networking, and this transport should only be selected if the image exposes OVF properties that bootstrap the guest's networking

akutz (changes requested):
Thanks @zyiyi11 , looks good aside from two nits.

api/v1alpha1/virtualmachine_types.go
zyiyi11: Thanks! I also made changes to all other comments on metadata transport to accommodate 80 columns.

akutz: Ha, thanks! I had done the same in a WIP branch for v1a2 which is why I just asked you to handle yours, but I should have known you would have not stopped at that :)

zyiyi11: haha yeah. I can't stand that they are not aligned. ;)

api/v1alpha1/virtualmachine_types.go
```diff
@@ -113,22 +113,38 @@ type VirtualMachineNetworkInterface struct {
...
 	VirtualMachineMetadataOvfEnvTransport VirtualMachineMetadataTransport = "OvfEnv"
 
+	// VirtualMachineMetadataVAppConfigTransport indicates that the data set in
+	// the VirtualMachineMetadata Transport Resource, i.e., a ConfigMap or Secret,
+	// will be vApp properties on the VM, which will be exposed as vAppConfig to the Guest VM.
+	// Only properties marked userConfigurable and already present in either
+	// OVF Properties of a VirtualMachineImage or as vApp properties on an existing VM
+	// or VMTX will be set, all others will be ignored.
+	//
+	// This transport should only be selected if the image exposes OVF properties
```
akutz, line 144: Maybe:

```go
// Selecting this transport means the guest's network is not automatically
// configured by vm-tools. This transport should be selected if the image
// exposes OVF/vApp properties that are used by the guest to bootstrap
// its networking configuration.
```

akutz (approved):

PR #102, opened 2022-08-29
Generate DeepCopy implementations for versioned templating

zyiyi11:
This change adds DeepCopy implementations for versioned templating data structures,
including: `VirtualMachineTemplate`, `NetworkStatus ` and `NetworkDeviceStatus`.


akutz (approved):
lgtm

PR #105, opened 2022-10-13
Change ProviderRef back

cyclexuxu:
The previous change which changed the `VirtualMachineImageSpec ProviderRef` from `ContentProviderReference` to `corev1.TypedLocalObjectReference`. That change breaks the current VM Operator codebase, specifically the `contentsource controller `. After discussion, we decided to change back for now. Once the new FSS is active, will continue to use ProviderRef to point to the Library Item.


akutz (approved):
Thank you!

PR #106, opened 2022-10-17
VM MinHardwareVersion

akutz:
This patch adds support for specifying a VM's mininum hardware version in order to enable more control over a VM's hardware version by the DevOps persona. This also enables upgrading brownfield VMs to a more recent hardware version in order to support features like PVCs and Dynamic Direct Path I/O devices.


api/v1alpha1/virtualmachine_types.go
```diff
@@ -354,6 +354,39 @@ type VirtualMachineSpec struct {
...
+	// It is important to remember that a VM's hardware version may not be
+	// downgraded and upgrading a VM deployed from an image based on an older
+	// hardware version to a more recent one may result in unpredictable
+	// behavior. In other words, please be careful when choosing to upgrade a
+	// VM to a newer hardware version.
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=13
+	// +kubebuilder:validation:Maximum=20
+	MinHardwareVersion int32 `json:"minHardwareVersion,omitempty"`
```
aruneshpa, line 389: Since this is an `int32` and is an `optional` field, do check against the value `0`, to determine that no minimum hardware version has been specified?  (As opposed to making this a pointer and checking against `null`).

akutz: Hmm, good question. When is the validation executed? I do not know. Brb 😃

akutz: It appears fine, at least via kubectl. I:

akutz: Updating `vm.yaml` to be the following:

```yaml
apiVersion: vmoperator.vmware.com/v1alpha1
kind: VirtualMachine
metadata:
  name: my-vm
spec:
  minHardwareVersion: 12
```

and reapplying it results in the expected error:

```bash
$ kubectl apply -f vm.yaml
The VirtualMachine "my-vm" is invalid: spec.minHardwareVersion: Invalid value: 12: spec.minHardwareVersion in body should be greater than or equal to 13
```

akutz: Anyway, I think this is fine. The entire purpose of `omitempty` is to *not* send a value if the value is empty. You make it `nil` when you want a type's empty value to be something that can be sent, for example, like when you _want_ to send a value of `0`. Since we never want to do that here, using `0` as the empty value should be fine. Clients should not send it, and so validation rules won't apply.

@vincepri , do you happen to know if the above is accurate? If `0` is the empty value, client-go won't send an actual `0`, right?

akutz: @aruneshpa, the `// +omitempty` indicates a field will not be serialized when it is the empty value. For integers this is `0`, so we're good.

aruneshpa: Got it.  Yeah, we should be good then.  
Another clarification is that even if this field is empty, because it is a primitive type, the controllers (when we read and de-serialize the object in its type) will still populate it as the default value of `0`, so unset doesn't really mean much here, right?

akutz: Yes, but we know zero isn't a valid hardware version, so that's not an issue. The only time it matter is if 0 *is* a valid value for some numeric field.

sreyasn:
Change looks good to me. We'll need to decide on what takes precedence when the VM Class Config Spec's hw version and the VM Spec's hw version are set. VM Spec overrides Class?.

sreyasn (approved):

akutz:
> Change looks good to me. We'll need to decide on what takes precedence when the VM Class Config Spec's hw version and the VM Spec's hw version are set. VM Spec overrides Class?.

I think we already made that decision when we said the hardware from the class overrides the hardware from the OVF 😃 

Thanks for the approvals -- I'm going to wait to merge this until we have go-ahead for Shruthi to execute on the implementation. We don't want to add the field until we know we can resource the work to make it available in the next release.

PR #107, opened 2022-10-17
Add more details to VirtualMachinePublishRequest status

yi0909:
This change includes
- Introduces a field Attempt to the VirtualMachinePublishRequestStatus. Attempt represents the number of times VM publish request has been attempted.
- Update VirtualMachinePublishRequestTarget comment.
- Add SourceRef and TargetRef in the VirtualMachinePublishRequestStatus.


api/v1alpha1/virtualmachinepublishrequest_types.go
```diff
@@ -215,6 +218,11 @@ type VirtualMachinePublishRequestStatus struct {
 	// +optional
 	StartTime metav1.Time `json:"startTime,omitempty"`
 
+	// Attempt represents the number of times the request has been attempted.
+	//
+	// +optional
+	Attempt int `json:"attempt,omitempty"`
```
akutz, line 224: Please change this to 

```golang
Attempts int `json:"attempts,omitempty"`
```

api/v1alpha1/virtualmachinepublishrequest_types.go
```diff
@@ -82,6 +82,10 @@ type VirtualMachinePublishRequestSource struct {
 type VirtualMachinePublishRequestTargetItem struct {
 	// Name is the name of the published object.
 	//
+	// If the spec.target.location.apiVersion equals imageregistry.vmware.com/v1alpha1 and
```
akutz, line 85: Please ensure the column length is <80 since all other comments in this file are.

akutz (changes requested):

api/v1alpha1/virtualmachinepublishrequest_types.go
yi0909: Changed. 👍 Also based on @aruneshpa 's suggestion in today's meeting, add a `lastAttemptTime` field to expose more details to users.

api/v1alpha1/virtualmachinepublishrequest_types.go
```diff
@@ -215,6 +234,11 @@ type VirtualMachinePublishRequestStatus struct {
 	// +optional
 	StartTime metav1.Time `json:"startTime,omitempty"`
 
+	// Attempt represents the information about the attempted requests.
+	//
+	// +optional
+	Attempts VirtualMachinePublishRequestAttempts `json:"attempt,omitempty"`
```
akutz, line 240: - Please update the JSON field name to `attempts`
- This should _probably_ be  type `int64` and tracking the last attempt time should likely occur as part of the conditions. Neither [PodStatus](https://pkg.go.dev/k8s.io/api/core/v1#PodStatus) or [ContainerStatus](https://pkg.go.dev/k8s.io/api/core/v1#ContainerStatus) tracks the last container-pull attempt time as part of the status, but rather as part of the events. Thus I do not know if there is precedent. I might even drop the `Attempts` entirely and simply emit the attempt count as an event to make this similar to pulling a container image.

akutz (changes requested):
Looking good. Can you please do a little digging on the precedent for how to track a counter in the status? See my in-line comment for context. Thanks!

api/v1alpha1/virtualmachinepublishrequest_types.go
akutz: If you do keep this, let's inline it directly into status, ex:

```golang
type VirtualMachinePublishRequestStatus struct {

    // ... ellided ...

    // Attempts represents the number of times the request has been attempted.
    //
    // +optional
    Attempts int64 `json:"attempts,omitempty"`

    // LastAttemptTime represents the time when the latest request was sent.
    //
    // +optional
    LastAttemptTime metav1.Time `json:"lastAttemptTime,omitempty"`
}
```

Still, I wonder if the last attempt time should be a condition?

yi0909: Synced up offline. Will keep `Attempts` and `lastAttemptTime` and incline them directly into status. Resolve this comment. 

`Attempts` are required as we need a field to generate different actIDs when sending publish requests. This field should not be immutable to users.

dilyar85 (approved):
LGTM.

api/v1alpha1/virtualmachinepublishrequest_types.go
```diff
@@ -179,7 +184,7 @@ type VirtualMachinePublishRequestSpec struct {
 	// will be marked in error.
 	//
 	// +optional
-	Target VirtualMachinePublishRequestTarget `json:"target,omitempty"`
+	Target *VirtualMachinePublishRequestTarget `json:"target,omitempty"`
```
aruneshpa, line 187: Out of curiosity, is the purpose of making these pointers so that you can check for `nil` values?  Even after marking these as `optional` and `omitempty`, do we still need these as pointers?

api/v1alpha1/virtualmachinepublishrequest_types.go
```diff
@@ -215,6 +220,16 @@ type VirtualMachinePublishRequestStatus struct {
 	// +optional
 	StartTime metav1.Time `json:"startTime,omitempty"`
 
+	// Attempts represents the number of times the request has been attempted.
```
aruneshpa, line 223: Maybe qualify the comment a little more?  Same comment for L231.
```suggestion
	// Attempts represents the number of times the request to publish the VM has been attempted.
```

api/v1alpha1/virtualmachinepublishrequest_types.go
akutz: I also prefer to avoid pointers wherever necessary as it avoids NPEs and is more efficient memory management. Yi is hitting a JSON patch error when these are not pointers, but frankly, we shouldn't be mutating the spec here anyway IMO. The intention was not to default via a mutation webhook.

cyclexuxu (approved):
LGTM

zyiyi11 (approved):

api/v1alpha1/virtualmachinepublishrequest_types.go
yi0909: Switched back to use struct.

akutz (approved):
lgtm, thank you again!

PR #109, opened 2022-10-20
Make ContentLibraryRef field be pointer

cyclexuxu:
This `ContentLibraryRef` field should be pointer, as it's optional and one of the fields in `TypedLocalObjectReference` is required  for third-party types.


yi0909 (approved):

munire0503:
@cyclexuxu Thanks for creating this PR. I have several changes in mind to add/update on this type. It would be nice if you can include them as well since you are touching this file already. I will send them out shortly in the following PR comment. Thanks!

dilyar85:
@cyclexuxu Thanks for creating this PR. I have several changes in mind to add/update on this type. It would be nice if you can include them as well since you are touching this file already. I will send them out shortly in the following PR comment. Thanks!

dilyar85:
@cyclexuxu Would you mind also updating the `printcolumn` of `ClusterVirtualMachineImage` type with the following changes, please?
- Rename `CONTENTSOURCENAME` to `ITEM-REF-NAME` (as we won't have `conentsource` type in the new image registry service)
- Rename `CONTENTLIBRARYNAME` to `CONTENT-LIBRARY-NAME` (with dashes for easier reading)
- Add `IMAGE-NAME` with `JSONPath=".status.imageName"`

cyclexuxu:
> @cyclexuxu Would you mind also updating the `printcolumn` of `ClusterVirtualMachineImage` type with the following changes, please?
> 
> * Rename `CONTENTSOURCENAME` to `ITEM-REF-NAME` (as we won't have `conentsource` type in the new image registry service)
> * Rename `CONTENTLIBRARYNAME` to `CONTENT-LIBRARY-NAME` (with dashes for easier reading)
> * Add `IMAGE-NAME` with `JSONPath=".status.imageName"` (this will display the vSphere image name while we encode the field name)

Hi @dilyar85 , thanks for the suggestion. I have updated the `CONTENTSOURCENAME` and `CONTENTLIBRARYNAME` as you suggested. Added the `ImageName`, but without the dash, as I think there already has `ImageSupported`, just follow the same pattern.

dilyar85:
> Hi @dilyar85 , thanks for the suggestion. I have updated the `CONTENTSOURCENAME` and `CONTENTLIBRARYNAME` as you suggested. Added the `ImageName`, but without the dash, as I think there already has `ImageSupported`, just follow the same pattern.

Yeah we should probably add a dash there too so that the column will be displayed as `IMAGE-SUPPORTED` (rather than the current `IMAGESUPPORTED` since it's not a single word). Could you please update all the multi-words columns ("OsType", "ImageSupported" and "ImageName") with a dash in between and move `Image-Name` to the top? Thanks.

yi0909 (approved):

yi0909:
Found that upstream uses space to separate words. e.g. `NOMINATED NODE` and `READINESS GATES` for pod.

cyclexuxu:
Changed to use space to separate non single word names as we can find same cases in upstream and other repos.

bryanv:
IDK if there is necessarily k8s convention around the column names. As you noted, pods use spaces, eg `NOMINATED NODE`, while nodes use dashes, eg `INTERNAL-IP`,`OS-IMAGE`.

My preference would be for dashes since that does make rudimentary CLI parsing easier (but there are safer & more stable ways to obtain those fields).

That said, it seems we should try at least keep ourselves (mostly) internally consistent: I think we only have one spot that uses spaces (`Content Library UUID`), one dash usage `PRIMARY-IP` (likely inspired by node), and then one word for everything else.

I think it would be best if we reduce the changes in this MR to just the purpose of the title, and have a separate discussion/MR to rename what needs to be renamed here and make other printer columns consistent with whatever we choose.

cyclexuxu:
> IDK if there is necessarily k8s convention around the column names. As you noted, pods use spaces, eg `NOMINATED NODE`, while nodes use dashes, eg `INTERNAL-IP`,`OS-IMAGE`.
> 
> My preference would be for dashes since that does make rudimentary CLI parsing easier (but there are safer & more stable ways to obtain those fields).
> 
> That said, it seems we should try at least keep ourselves (mostly) internally consistent: I think we only have one spot that uses spaces (`Content Library UUID`), one dash usage `PRIMARY-IP` (likely inspired by node), and then one word for everything else.
> 
> I think it would be best if we reduce the changes in this MR to just the purpose of the title, and have a separate discussion/MR to rename what needs to be renamed here and make other printer columns consistent with whatever we choose.

Thanks. I will move the renaming part to another MR. CC @dilyar85

zyiyi11 (approved):

PR #110, opened 2022-10-24
Add VirtualMachinePublishRequest reason.

yi0909:
This change includes
- Introduce VirtualMachinePublishRequest reason.
- Update MarkCondition function. It should set the reason to the input parameter if it is not empty.


api/v1alpha1/condition_consts.go
```diff
@@ -106,3 +106,33 @@ const (
...
 )
+
+// Condition.Reason for Conditions related to VirtualMachinePublishRequest.
+const (
+	// VirtualMachinePublishRequestSourceVMNotFoundReason (Severity=Error) documents that the source VM is not found.
+	VirtualMachinePublishRequestSourceVMNotFoundReason = "SourceVMNotFound"
+
+	// VirtualMachinePublishRequestSourceVMNotCreatedReason (Severity=Error) documents that the source VM
+	// hasn't been fully created yet.
+	VirtualMachinePublishRequestSourceVMNotCreatedReason = "SourceVMNotCreated"
```
bryanv, line 117: Do we have sufficient context to determine between this reason and the UniqueIDNotReady below? Once the VM is created, we have the MoID/UniqueID. What does "fully created" mean in the context here?

PR #111, opened 2022-10-26
Update VirtualMachineImage related fields and separate multi-words with dash in print column

dilyar85:
This PR introduces the following changes:
- Add an optional `ContentVersion` field in `VirtualMachineImageStatus`
- Add some constants related to `VirtualMachineImage` conditions
- Update `printcolumn` remarks by adding a dash in multi-word access all types to keep consistency

### DIFF
--- a/api/v1alpha1/condition_consts.go
+++ b/api/v1alpha1/condition_consts.go
@@ -93,6 +93,12 @@ const (
 	// VirtualMachineImageV1Alpha1CompatibleCondition denotes image compatibility with VMService. VMService expects
 	// VirtualMachineImage to be prepared by VMware specifically for VMService v1alpha1.
 	VirtualMachineImageV1Alpha1CompatibleCondition ConditionType = "VirtualMachineImageV1Alpha1Compatible"
+
+	// VirtualMachineImageSyncedCondition denotes that the image is synced with the vSphere content library.
+	VirtualMachineImageSyncedCondition ConditionType = "VirtualMachineImageSynced"
+
+	// VirtualMachineImageProviderReadyCondition denotes that the image provider from vSphere is ready.
+	VirtualMachineImageProviderReadyCondition ConditionType = "VirtualMachineImageProviderReady"
 )
 
 // Condition.Reason for Conditions related to VirtualMachineImages.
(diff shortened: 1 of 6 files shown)


zyiyi11 (approved):
LGTM

api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -175,12 +176,13 @@ func (clusterVirtualMachineImage *ClusterVirtualMachineImage) SetConditions(cond
 // +kubebuilder:resource:scope=Cluster,shortName=cvmi;cvmimage;clustervmimage
 // +kubebuilder:storageversion
 // +kubebuilder:subresource:status
-// +kubebuilder:printcolumn:name="ContentSourceName",type="string",JSONPath=".spec.providerRef.name"
-// +kubebuilder:printcolumn:name="ContentLibraryName",type="string",JSONPath=".status.contentLibraryRef.name"
+// +kubebuilder:printcolumn:name="Provider-Name",type="string",JSONPath=".spec.providerRef.name"
```
ericvmw, line 179: Just curious on what can be the provider-name of an VM image, any example?
Also is the new printcolumn naming pattern (separated by -) consistent with other APIs?

api/v1alpha1/condition_consts.go
```diff
@@ -105,4 +111,12 @@ const (
 	// VirtualMachineImageV1Alpha1NotCompatibleReason (Severity=Error) documents that the VirtualMachine Image
 	// is not prepared for VMService consumption.
 	VirtualMachineImageV1Alpha1NotCompatibleReason = "VirtualMachineImageV1Alpha1NotCompatible"
+
+	// VirtualMachineImageProviderNotReadyReason (Severity=Error) documents that the VirtualMachine Image provider
+	// from the vSphere content library is not ready.
```
ericvmw, line 116: Is the provider here the `vSphere content library` or some provider from vSphere content library? just to understand what this provider means here.

api/v1alpha1/virtualmachineimage_types.go
dilyar85: @ericvmw "provider-name" is the name of content-library-item (or cluster-content-library-item).  I used this name here to match the JSON path name of the resource field.
There is no strict convention on printcolumn naming patterns from K8s (it uses both "-" and space for the internal resources). I chose "-" as suggested by BV' [comment](https://github.com/vmware-tanzu/vm-operator-api/pull/109#issuecomment-1289678131) and updated all our APIs in this PR to keep the consistency.

ericvmw: I see. So this is the resource name of the corresponding content-library-item or cluster-content-library-item, and the `image-name` in the status is the name of the item / image from vSphere.

ericvmw (approved):

PR #112, opened 2022-10-31
Add OvfProp Desc & Label fields

akutz:
This patch adds the Description and Label fields to an OVF property for a VirtualMachineImage.


zyiyi11 (approved):
LGTM

